PR #41, opened 2017-05-14
Added link to emacs package "flycheck-vale".

abingham:
I added the link in alphabetical order with the rest, but let me know if it should just go at the end instead. 

And thanks, vale is brilliant!


jdkato (approved):
Looks great&mdash;thanks!

PR #61, opened 2018-03-05
[WIP] Added Google style (based on Google style guide for developers)

gaurav-nelson:
@jdkato Can you check this one?

based on https://developers.google.com/style/


jdkato:
It looks great!

In fact, I've been looking for a thorough, non-commercial style guide to showcase in a "boilerplate" repository&mdash;and Google's guide appears to be a good fit for this.

The idea would be to provide a basic structure (e.g., unit tests and a CI set up) for future style creators to use without forcing them to navigate Vale's actual code base. I've put together a [simple example ](https://github.com/ValeLint/vale-boilerplate) that uses [Cucumber](https://cucumber.io/) + Travis CI for its tests.

What do you think of this idea?

gaurav-nelson:
Thanks @jdkato 
Keeping styles separated does make sense. But I do not understand how this will work with vale? will each user fork the vale-boilerplate repo and create there own styles with it? where will they publish?

What do you think about having a separate space for Styles, like `ValeLint/Styles` for the styles directory, with tests built in like the vale-boilerplate example. So the style creators can fork the repo, do there additions, then create PR to get their work in. This will help in keeping all the styles accessible from a single location.

jdkato:
My vision for v1.0.0 of Vale is this:

1. Vale will ship with a collection of general-purpose style building blocks such as rules for passive voice, gender bias, and technical terminology (e.g., "JavaScript", not "Javascript").
2. Individual users will then use these building blocks to make a house style that they host and contribute to on their own (perhaps starting from `vale-boilerplate`). This allows them to do things like incorporate brand-specific terminology and maintain their own spelling dictionaries (see, for example, [Write the Docs](https://github.com/writethedocs/www/blob/master/vale/WTD/Branding.yml) and [Linode](https://github.com/linode/docs/blob/master/ci/vale/)).

And while I can see the appeal of pre-made styles, it just doesn't seem like they're used in practice. I've yet to find a project/organization that adheres to a non-trivial style guide that  they didn't make themselves&mdash;e.g., to name a few, [Mapbox](https://blog.mapbox.com/regulating-english-with-retext-mapbox-standard-d79a8158f251), [Shopify](https://github.com/Shopify/retext-shopify), [MailChimp](https://styleguide.mailchimp.com/), [Homebrew](https://docs.brew.sh/Prose-Style-Guidelines.html), and [GCC](https://gcc.gnu.org/codingconventions.html#Spelling) all have their own.

Have you had a different experience?

gaurav-nelson:
Thanks @jdkato for sharing your vision. Yes, I agree, pre-made styles are not used in practice.

gaurav-nelson:
In that case do you want me to close this PR and use the boilerplate to publish that style?

jdkato:
Sure. 

I'll work on writing a better README for the boilerplate. Also, to avoid duplicating your work on Google's guide, I'll change the boilerplate's example style to [18F](https://content-guide.18f.gov/).

gaurav-nelson:
Thank you @jdkato , I am closing this one now.

gaurav-nelson:
@jdkato Thanks for providing the boilerplate. I am not sure how to setup features and fixtures for it. Can I get one done as an example in https://github.com/gaurav-nelson/vale-google, A PR would be awesome, or if you can point me to some resources to checkout on how to do that. Thanks again 👍

PR #75, opened 2018-08-07
Update Quotes.yml

ChrisChinchilla:
Allow for punctuation check in single and double quotes

### DIFF
--- a/styles/18F/Quotes.yml
+++ b/styles/18F/Quotes.yml
@@ -4,3 +4,4 @@ level: error
 nonword: true
 tokens:
   - '"[^"]+"[.,?]'
+  - "'[^']+'[.,?]"


jdkato:
Due to the explanation in #78, I'll pass on merging this.

If you'd like to discuss improving 18F specifically, you can open an issue on the [boilerplate](https://github.com/errata-ai/vale-boilerplate). Of course you can also share your thoughts on my v1.0.0 plan over on #78.

Thanks!

ChrisChinchilla:
OK @jdkatoI see now, this is what the style guide specifies, which makes it somewhat unuseful to technical writers (or others who use single quotes sometimes), but fair enough, this matches this style :)

PR #76, opened 2018-08-07
Update So.yml

ChrisChinchilla:
Change rule so it also matches 'So' with a comma, i.e.

`So` and `So,`

### DIFF
--- a/rule/write-good/So.yml
+++ b/rule/write-good/So.yml
@@ -3,4 +3,4 @@ message: "Don't start a sentence with '%s'"
 ignorecase: true
 level: error
 raw:
-  - '(?:^[^\w]*|[;-]\s)(so)\s'
+  - '(?:^[^\w]*|[;-]\s)(so)'


jdkato:
Thanks!

However, `(?:^[^\w]*|[;-]\s)(so)` will fail `write-good`'s test suite (see [here](https://github.com/btford/write-good/blob/1a32ec6309352ea814e77365bdbe382c2a9ddf08/test/basic.spec.js#L65-L68) and [here](https://regex101.com/r/SFNqDs/1)). `(?:^[^\w]*|[;-]\s)(so)[\s,]` seems to be more inline with `write-good`'s functionality.

In 402d5559d7a5c777f5e9f2f31fc797f64398186d and 47f85eeb6971a3d976bdab64128cb1cbb7bfaaf5, I've updated the defintion and made it so Travis CI will catch failures related to [`rule.go`](https://github.com/errata-ai/vale/blob/master/rule/rule.go) in the future.

ChrisChinchilla:
Thanks @jdkato I am a regexp n00b :/

Issue #26, opened 2017-03-14
Tests? Possible issue with GenderBias


plaindocs:
Are there some tests I can run that'll convince me all checks are working? :-p

I'm currently manually testing GenderBias and it does not seem to be working perfectly.

I does not trigger on `doorman` but should. (Also typo in concierge.)

Thanks :-)

jdkato:
You can run `make test`, which runs the .feature files (see the `features` directory). This requires you to have Go + Ruby installed and configured, though. There are currently 43 scenarios and 148 steps in [Cucumber](https://cucumber.io) terms.

It looks like `doorman` isn't flagged due to a missing "n" in its definition file (fixed in 1421434223f9edaa06d621692b7d8c7fa630d690).

Going forward:

- I'm working on re-writing the limited documentation I have (most will be at the [docs](https://github.com/ValeLint/docs) repo), which will include a guide on setting up a development environment.
- As I'm documenting Vale's features, I'll also be expanding its test suite (my goal is to be able to remove the "pre-release" label by v0.5.0).
- I'm still working out what role Vale should play in testing external styles. (i.e., Should it have some functionality built in or should that be left up to the users?)

plaindocs:
Ha, can't believe I saw the other typo but not that one.

I'll take a look at the docs repo when I've got the time, instead of the wiki - or let me know if you'd like help on something specific there.

jdkato:
I'm closing this since the typo has been fixed and, in addition to adding 23 steps to the test suite, I've started:

1. Testing all regular expressions used in Vale's built-in rules to ensure that they compile and match their intended tokens (see [lint_test.go](https://github.com/ValeLint/vale/blob/master/lint/lint_test.go)).
2. Adding a simple once-through test case for each rule (see 20c2fd6db5d5b2e77fcd10455f8ace4a8a6143a2).

Also, thanks for the offer on the docs. I'm trying to update it at least once a week, and I'll let you know if I come across something specific that I'm having a hard time explaining.

plaindocs:
To test it properly shouldn't we be taking the regexes directly from https://github.com/ValeLint/vale/blob/master/rule/GenderBias.yml instead of duplicating them?

jdkato:
I haven't finished (1) or (2) from above (notably, I need to add PassiveVoice to `lint_test`), but in general I think the two serve different purposes (although a malformed regex would show up in both):

[...]

PR #89, opened 2018-10-25
Fix typo convierge -> concierge

wouter-veeken:


jdkato:
Thanks -- I should've actually fixed this in https://github.com/errata-ai/vale/issues/26, but it appears that "convierge" is ignored by [misspell](https://github.com/client9/misspell).

Perhaps it's time for Vale to spell-check itself ...

PR #90, opened 2018-10-25
Make write-good.Passive message more meaningful

wouter-veeken:
There was no mention of passive or active voice in the original message. I think my version is a little more helpful.


jdkato:
Thanks!

PR #108, opened 2019-02-02
Add link to Neomake support in Vim as well

decentral1se:
Available in https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/markdown.vim.


jdkato:
Thanks!

PR #111, opened 2019-02-28
fix: broken links

svx:
This PR fixes two broken links.
The links are pointing now to https://github.com/testthedocs/vale-styles/tree/master/18F.


jdkato:
Thanks!

PR #121, opened 2019-04-18
docs: Fix a broken link

felicitymay:
This link had a double set of parentheses which stopped it working correctly.

I've read the Contributing guidelines and hope that I'm raising this with the appropriate commit comments etc. Please do correct me if I've made any errors.


jdkato:
It looks good -- thanks!

PR #166, opened 2020-01-09
Fix doc link for the Docker image

rfdonnelly:
The previous link (cloud.docker.com) required authentication.  The new link (hub.docker.com) does not.

Also fix terminology.  Change "container" to "image".  A Docker container is an instance of a Docker image.  Docker Hub hosts Docker images.  The term "container image" is also valid.


jdkato:
Thanks!

PR #187, opened 2020-04-10
[WIP] Add draft of Flare guidance

ChrisChinchilla:
@jdkato I didn't really know how best to start progress on some of the research we discussed, and I leave it up to you if it's best to track with issues, or projects, or something else. I also think in the long run, any _actual_ support should be added to Vale server, but for now, advice in the CLI tool docs I think is a good start.

As a follow up issue to track, Flare does offer plugin architecture, there's a link in the PR, I didn't dig much further yet.


jdkato:
My inclination here is to simply not cover the topic until we have something more "meaningful" (i.e., other then "we're looking into it") to say. I'm close to having a working extension for Confluence, so that could be a good starting point for a commercial tools section.

ChrisChinchilla:
@jdkato Hmm, having been asked this question a lot, how about instead, in the meantime I change this to a more broad "FAQ" on "Does vale support tool x"

jdkato:
Sure -- I'll admit that I didn't realize that this topic came up regularly (I don't think I've ever been personally asked about it) but, since it does, it's indeed probably worth clarifying until there's more concrete work to mention.

ChrisChinchilla:
@jdkato refactored to FAQ format. I'd say merge what we have, and add new tools as we research them, which I am doing slowly.

PR #221, opened 2020-07-30
style: suggest non-oppressive terminology

clywyk:
I'm a new user to Vale and appreciate, for example, your suggestions for inclusive language. As a result, when I saw `whitelist` and `blacklist` when running `vale ls-config, I was surprised. I'm not sure what all you'll need to do from the backend to make this shift, but consider using something like `allowlist/blocklist` or `permit/block` instead.

Here's a good resource: https://tools.ietf.org/id/draft-knodel-terminology-00.html#rfc.section.1.2

Have a great day!

### DIFF
--- a/check/check.go
+++ b/check/check.go
@@ -895,7 +895,7 @@ func (mgr *Manager) loadStyles(styles []string, loaded []string) []string {
 }
 
 func (mgr *Manager) loadVocabRules(config *core.Config) {
-	// Whitelist
+	// Allowlist or Permit
 	if len(config.AcceptedTokens) > 0 {
 		vocab := Substitution{}
 		vocab.Extends = "substitution"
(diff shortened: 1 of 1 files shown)


jdkato:
This has already been changed (besides these comments) in the [latest version](https://github.com/errata-ai/vale/releases/tag/v2.3.0-beta): you'll no longer explicitly see any mention of these data structures in the output of `ls-config` and internally they're referred to as `AcceptedTokens` and `RejectedTokens`.

Thanks!

ChrisChinchilla:
Thanks for the contribution @clywyk ! This also prompted me to think that while Vale can be used to create checks for inclusive language, it might be good to make that more specific, or write a specific blog post about the topic.

clywyk:
Thanks to you both! Keep up the good work. :) @jdkato @ChrisChinchilla

PR #234, opened 2020-09-15
Add support for org mode markup

dmacvicar:
Uses the [go-org](https://github.com/niklasfasching/go-org/org) library and convert to HTML, just like the markdown implementation.

Screenshot from 2020-09-15 21-50-18

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 Vale is a command-line tool that brings code-like linting to prose. It's cross-platform (Windows, macOS, and Linux), written in Go, and released under the MIT license. Some of its key features are:
 
-* [Usability](https://docs.errata.ai/vale/scoping): Vale supports HTML, Markdown, AsciiDoc, reStructuredText, XML, and
+* [Usability](https://docs.errata.ai/vale/scoping): Vale supports HTML, Markdown, Emacs Org mode, AsciiDoc, reStructuredText, XML, and
   DITA&mdash;allowing it to "just work" on *real-world* content.
 
 * [Extensibility](https://docs.errata.ai/vale/styles): Unlike most writing-related software, Vale's primary purpose isn't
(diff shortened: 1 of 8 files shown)


jdkato:
Thanks for the PR.

Could you add a format test case? There are two steps to adding a test case:

1. Add a file, `test.org`, to the [format fixtures](https://github.com/errata-ai/vale/tree/master/fixtures/formats). You can use the [Markdown test](https://github.com/errata-ai/vale/blob/master/fixtures/formats/test.md) as an example: the idea is to ensure that different parts of the format are handled correctly (code blocks, URLs, etc.). The only enabled rule looks for annotations (`NOTE`, `TODO`, `XXX`, and `FIXME`), which can be added throughout the file.

2. Add a new [Cucumber Step](https://github.com/errata-ai/vale/blob/master/features/lint.feature) like the one for Markdown (shown below) that has the expected output of running Vale on the test file added in the first step:

    https://github.com/errata-ai/vale/blob/9d551c8fb0edf6efd6ffe7821130018f2c8f1f00/features/lint.feature#L97-L111

dmacvicar:
Thanks. I totally missed those tests.

I will dig into `lintHtml` in order to be able to pass those. Eg. org mode `TODO` in a heading is a agenda TODO and should not be part of _Annotations_ warnings.

jdkato:
The annotations part of the tests is actually somewhat arbitrary; it could easily use another rule, so I wouldn't worry too much about that. 

The idea is to merely ensure that Vale is handling the format correctly -- in Markdown, for example, this means that no alerts should be raised for text in code spans, code blocks, or URLs.

I think the best first step here is to simply add a representative example file for org mode. We can then discuss what should or shouldn't raise an alert.

dmacvicar:
I will send a new PR when I get time again to work on this.

fpatz:
@dmacvicar, any plans to get back to this one day? I'm interested in getting org-mode supported in vale and would be willing to work on this (e.g. by adding the test case).

dmacvicar:
@fpatz I'd love to.

I rebased my changes and tried to get a dev environment again.

Are you able to get the cucumber testsuite to pass locally on master? I can't get it to pass here. Scenarios like "Sequence" get no output.

tecosaur:
Hi @dmacvicar, what would you say the chances of you getting back to this at some point might be?

PR #279, opened 2020-12-02
Feature: Add support for outer.control in DITA

pklaschka:
### Disclaimer
I am by no means a Go developer (I'm primarily a full-stack web developer with some minimal C++ and Java experience in addition to that, Makefiles, etc. aren't really in my repetoire 😉). Therefore, finding the correct position in the source code and a way to implement it (I have a feeling this is probably the worst piece of code I've ever submitted as a Pull Request) already was interesting enough. I wasn't able (and currently do not have the time to learn how) to set up and write for all the linters, testing systems, etc. While the change was rather small and should fully match the previous code style (as far as my understanding of Go goes), this PR *just* includes the source code changes. Maybe we can find a way to include it, sooner or later, anyway?

### Description
Previously, DITA files using something like `@conref` to a parent directory lead to an error as `outer.control` was not allowed in the internal call of the DITA command. This is now supported. Some additional code changes were required to find the generated .html file as this means that the .html file can now be in a subdirectory of the temporary folder.

### Example use case
Consider the following example:

#### `references/a.dita`
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE reference PUBLIC "-//OASIS//DTD DITA Reference//EN" "reference.dtd">
<reference id="key_git_principles">
    <title>Reference A</title>
    <shortdesc><ph conref="../reusables/phrases.dita#phrases/ph_gxh_bcv_vnb"/></shortdesc>
</reference>
```

#### `reusables/phrases.dita`
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd">
<topic id="phrases">
    <title>phrases</title>
    <body>
        <section id="section_psc_4bv_vnb">
            <title>Reusable Phrases part 1</title>
            <p>
                <ph id="ph_gxh_bcv_vnb">example-phrase</ph>
            </p>
        </section>
    </body>
</topic>
```

Previously, linting `references/a.dita` using vale would have been impossible as `outer.control` was set to fail (and subdirectory structures coming with it were unsupported).

With this PR, this can successfully get linted.

### DIFF
--- a/lint/markup.go
+++ b/lint/markup.go
@@ -329,17 +329,29 @@ func (l Linter) lintDITA(file *core.File) error {
 		"-o",
 		tempDir,
 		"--nav-toc=none",
-		"--outer.control=fail"}...)
+		"--outer.control=warn", // allows DITA files to reference external files, like in conrefs.
+	}...)
 	cmd.Stderr = &out
 
 	if err := cmd.Run(); err != nil {
 		return core.NewE100(file.Path, errors.New(out.String()))
 	}
 
-	basename := filepath.Base(file.Path)
-	data, err := ioutil.ReadFile(filepath.Join(
-		tempDir,
-		strings.TrimSuffix(basename, filepath.Ext(basename))+".html"))
+	var htmlFile string
+
+	targetFileName := strings.TrimSuffix(filepath.Base(file.Path), filepath.Ext(file.Path)) + ".html"
+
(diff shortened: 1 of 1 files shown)


jdkato:
Thanks -- this looks good!

These changes should be in a new release soon.

jdkato:
This is available now in v2.6.7.

Issue #297, opened 2020-12-23
Vale should ignore xref target (even when the macro has no attribute)


ggrossetie:
- OS: Linux 
- Install method: Direct download
- Vale version: `vale version 2.6.8`

`.vale.ini`

```ini
StylesPath = docs/.vale
MinAlertLevel = suggestion

[*.adoc]
BasedOnStyles = test
```

`docs/.vale/test/Substitutions.yml`

```yml
extends: substitution
message: Consider using '%s' instead of '%s'
level: error
ignorecase: true
swap:
  ruby: Ruby
```

`docs/test.adoc`

```adoc
= Document Title

Head over to xref:ruby.adoc[].
```

#### Current result

```console
$ vale docs/test.adoc

 docs/test.adoc
 3:19  error  Consider using 'Ruby' instead   test.Substitutions 
              of 'ruby
```

#### Expected result

I would expect the target of the `xref` to be ignored by Vale.

`xref` does not necessarily have attribute. In some context, the xref text can be inferred so it's actually a good practice to not explicitly define a cross reference text.

jdkato:
See #298.

ggrossetie:
I see, Vale is working on the generated HTML so it does not know that the cross-reference does not have an associated text.

The problem is that the built-in HTML 5 converter does not make any distinct so I cannot use `IgnoredClasses`. For reference, the generated HTML will be:

```html
<a href="ruby.adoc">ruby.adoc</a>
```

One way to solve this issue would be to provide a custom converter where we could add a class when the xref does not have any associated text.

I understand that Vale is not aware of the `xref` macro but I still think we should try to find a solution in order to provide a great user experience when working with AsciiDoc.

jdkato:
I think the problem here is not that the cross-reference lacks associated text; it's that its default associated text is violating the provided rule.

To me, this seems like a fair (and consistent) result from Vale: in this case, the text "ruby.adoc" is indeed both reader-facing and a rule violation.

A reasonable solution here would be to provide label text that doesn't violate the style guide.

Similarly, while you *could* use `IgnoredClasses` in #298, I would argue that a better solution would be to not use a bare link.

ggrossetie:
Actually, the associated text will be resolved by the site generator. In this case, it will use the document title of the file ruby.adoc.

My point is that Vale should not apply rules on the URL of the linked resource.

What about ignoring the text of the link if the value is equals to the `href` attribute? I think that's what most people will expect.
[...]

PR #301, opened 2020-12-29
Ignore link text when the text is equals to href

ggrossetie:
Ignore the text of a link when the value is equals to the href attribute. As a result, the linter will ignore the following occurrences:

```html
<a href="https://www.google.fr">https://www.google.fr</a>
```

```html
<a href="file.html">file.html</a>
```

In the above case, I believe that Vale should not lint the text since the value is actually an URL. Vale will still lint the text when the value is not equals to the href. For instance:

```html
<a href="https://www.google.fr">www.google.fr</a>
```

```html
<a href="foo.html">bar.html</a>
```

Using [`url.Parse`](https://golang.org/pkg/net/url/#Parse) to check that the text is an URL could produce false positive.

I think this change is reasonable since we know that the text of the link is actually an URL (since the value is equals to the `href` attribute). As mentioned in #297, it should be possible to define a `BlockIgnores` to explicitly ignore bare links but I think it should be built-in.

ref #297

### DIFF
--- a/features/lint.feature
+++ b/features/lint.feature
@@ -51,6 +51,7 @@ Feature: Lint
       """
       test2.adoc:3:17:Test.Rule2:Consider using 'AsciiDoc' instead of 'Asciidoc'
       test2.adoc:11:7:Test.Rule2:Consider using 'AsciiDoc' instead of 'asciidoc'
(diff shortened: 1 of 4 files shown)


jdkato:
Please see https://github.com/errata-ai/vale/issues/297#issuecomment-752210686.

Issue #365, opened 2021-09-20
Support code-climate output


axilleas:
:wave: it would be great if support for [code-climate](https://github.com/codeclimate/codeclimate) reports was added in the [`--output` option](https://docs.errata.ai/vale/cli#--output)!

Sure, you could do this using a template, but it would be nice if this was a built-in option. I see https://github.com/golangci/golangci-lint has this option, so you might get an idea how this is implemented.

My current hack is to use a template:

```golang
{{- /* Modify Vale's output https://docs.errata.ai/vale/cli#--output */ -}}

{{- /* Keep track of our various counts */ -}}

{{- $e := 0 -}}
{{- $w := 0 -}}
{{- $s := 0 -}}
{{- $f := 0 -}}

{{- /* Range over the linted files */ -}}

{{- range .Files}}

{{- $f = add1 $f -}}
{{- $path := .Path -}}

{{- /* Range over the file's alerts */ -}}
[

{{- range .Alerts -}}

{{- $error := "" -}}
{{- if eq .Severity "error" -}}
    {{- $error = .Severity -}}
    {{- $e = add1 $e  -}}
{{- else if eq .Severity "warning" -}}
    {{- $error = .Severity -}}
    {{- $w = add1 $w -}}
{{- else -}}
    {{- $error = .Severity -}}
    {{- $s = add1 $s -}}
{{- end}}

{{- /* Variables setup */ -}}

{{- $path = $path -}}
{{- $loc := printf "%d" .Line -}}
{{- $check := printf "%s" .Check -}}
{{- $message := printf "%s" .Message -}}
{{- $link := printf "%s" .Link -}}

{{- /* Output */ -}}

  {
    "description": "{{ $message }}",
    "fingerprint": "{{ $path }}-{{ $loc }}",
    "severity": "{{ $error }}",
    "location": {
      "path": "{{ $path }}",
      "lines": {
        "begin": {{ $loc }}
      }
    }
  },
{{end -}}
{{end -}}
]
```

I don't know how to omit the last comma for the last object though :thinking:

jdkato:
You can omit the last comma by delaying the placement with a loop counter (e.g., `{{- range $idx, $a := .Alerts -}}`):

<details>

```golang
{{- /* Modify Vale's output https://docs.errata.ai/vale/cli#--output */ -}}

{{- /* Keep track of our various counts */ -}}

{{- $e := 0 -}}
{{- $w := 0 -}}
{{- $s := 0 -}}
[...]

Issue #350, opened 2021-06-12
Add severity to `vale --output=line` output


guillaume-d:
OK, I know for special needs `--output=JSON` and Go-template-based output-styles (see below) exist already, but maybe this output style may still be expanded.

For example currently one gets:

```
README.md:72:71:write-good.Weasel:'likely' is a weasel word!
```

More useful would be:

```
README.md:72:71: [warning] write-good.Weasel: 'likely' is a weasel word!
```

Or for better backwards-compatibility (the message is the last part and not structured => unlikely to be parsed):

```
README.md:72:71:write-good.Weasel: [warning] 'likely' is a weasel word!
```
----
A workaround is indeed possible using a Go template file:
```
{{- /* Very much like `vale --output=line`, but with the severities too! -}}
{{- /* See also https://docs.errata.ai/vale/cli#template-examples */ -}}
{{- range .Files}}
{{- $p := .Path -}}
{{- range .Alerts -}}
{{- printf "%s:%d:%d: [%s] %s: %s" $p .Line (index .Span 0) .Severity .Check .Message }}
{{end -}}
{{end}}
```

jdkato:
Using a template file isn't a "workaround" here -- it's the intended solution.

Issue #260, opened 2020-11-02
Feature request: Include link in output


Ravlen:
As per the first example you see at https://docs.errata.ai/vale/styles, we can include a `link: ` entry in a style YAML. I think it would be extremely useful to have an option to enable the output of that link in error messages.

So, for example, using our [Alert Box rule](https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/.vale/gitlab/AlertBoxStyle.yml):

```
 README.md
 21:1  error  Alert box " NOTE: **Note:** This alert box is not following standards." must use the formatting in the style guide.  gitlab.AlertBoxStyle 

✖ 1 error, 0 warnings and 0 suggestions in 1 file.
```

Would instead look like:

```
 README.md
 21:1  error  Alert box " NOTE: **Note:** This alert box is not following standards." must use the formatting in the style guide.  gitlab.AlertBoxStyle 
 Link: https://docs.gitlab.com/ee/development/documentation/styleguide.html#alert-boxes

✖ 1 error, 0 warnings and 0 suggestions in 1 file.
```

This could be turned on and off within individual rules, but I think the minimal solution would be to turn it on and off globally in the config file. Something like `OutputStyleLinks = true`?

Ravlen:
@jdkato WDYT? I think this suggestion would be useful for everyone, not just GitLab 😅

jdkato:
There have been a number of requests to make minor adjustments to existing output formats and my response has typically been to repurpose the JSON output -- such as shown in https://github.com/errata-ai/vale/issues/232#issuecomment-685836520.

However, I've recently had the idea of introducing an `--output=custom` option that makes uses of Go's [template](https://golang.org/pkg/text/template/) library. 

I think this will address all of these issues in a fairly straightforward and useful way: You'll have template-level access to the entire slice of Alerts while being able to make use of the built-in utility functions (such as the coloring, underlining, etc.).

jdkato:
Custom output formats are now available on master, but I'd like to dicuss the implementation a little before releasing. The syntax will problably seem familiar if you've used Hugo (or Go templating in general).

- The format requested in #232 could now be implemented with the following template (`line.tmpl`):
 
    ```
    {{- range . -}}
      {{- .Path}}:{{.Line}}:{{index .Span 0}}:{{.Check}}:{{.Severity}}:{{.Message}}
    {{end -}}
    ```
    ```console
    $ vale --output='line.tmpl' file.md
[...]

Issue #232, opened 2020-09-02
Include severity in `--output line`


nostodnayr:
The line output is great because it's easy to parse, but you lose the message severity which makes it considerably less useful. 

The only way I can see around it is to add the severity to the filename for the rule. It's clunky, and would only work for first-party rules without trouble.

Right now I'm working with BBEdit's `bbresults` tool, but we'll use similar techniques when we deploy this in our CI/CD environment.

nostodnayr:
I modified the output and built a custom version of vale and at least when working with BBEdit, I find this output most useful: 

`file:line:column:check:severity:message`

I had `(check)` at the end of the message, but I expect for writers knowing this isn't that useful. It's only helpful now as I'm debugging. I would probably discard the `check` in most cases.

If you wanted to stay closer to the UNIX 'standard', `file:line:column:check:severity:message`, where a 'dumb' parser may reasonably expect to match any character that follows the colon after `check`. If you know what you're looking for, though, you'll get the richer information.

jdkato:
Using [`--no-wrap`](https://docs.errata.ai/vale/cli#--no-wrap) should make the default output just as easy to parse.

nostodnayr:
> Using [`--no-wrap`](https://docs.errata.ai/vale/cli#--no-wrap) should make the default output just as easy to parse.

True, and I've tried that, but the file is only printed once in that case, which doesn't agree with `bbresults` certainly.

jdkato:
I can't say that I'm overly familiar with `bbresults`, but I think it's fair to say that:

1. Altering `--output=line` would constitute a breaking change, which means it'd probably be better to just add a new output format; and

2. it's not reasonable to expect Vale to personally accommodate every other tool -- especially when they're so limited in what they expect and Vale already offers multiple options that provide the required information.

For example, you can get your required format trivially by writing a simple wrapper using your favorite scripting language (Python shown below):

```python
import sys
import json

data = json.load(sys.stdin)
for src, alerts in data.items():
    for a in alerts:
        print('{0}:{1}:{2}:{3}:{4}:{5}'.format(
            src,
            a['Line'],
            a['Span'][0],
            a['Check'],
            a['Severity'],
            a['Message']))
```

```console
$ vale --output=JSON fix.md | python bbresults.py 
fix.md:3:13:Microsoft.Vocab:suggestion:Verify your use of 'allows' with the A-Z word list.
fix.md:3:27:Microsoft.Adverbs:warning:Consider removing 'easily'.
[...]

PR #379, opened 2021-11-06
feat: Support Severity in line output

rockerBOO:
This adds severity to the line output. 

```
.../hybrids.md:25:270:warning:Google.Will:Avoid using 'will'.
.../hybrids.md:25:291:error:Vale.Spelling:Did you really mean 'squrriels'?
```

Previous

```
.../hybrids.md:25:270:Google.Will:Avoid using 'will'.
.../hybrids.md:25:291:Vale.Spelling:Did you really mean 'squrriels'?
```

This **will break** support for previous matches so maybe a better approach to adding this to the line output. Maybe adding another version of the line output for backwards compat. 

```
vale --output=line-severity
```

### DIFF
--- a/internal/cli/line.go
+++ b/internal/cli/line.go
@@ -33,8 +33,8 @@ func PrintLineAlerts(linted []*core.File, relative bool) bool {
 			if a.Severity == "error" {
 				alertCount++
 			}
-			fmt.Print(fmt.Sprintf("%s:%d:%d:%s:%s\n",
-				base, a.Line, a.Span[0], a.Check, a.Message))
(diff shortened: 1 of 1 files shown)


jdkato:
Yes, introducing a breaking change for an output format is a non-starter for me.

Additionally, this particular issue has been discussed multiple times (#232, #260, #350, #365, and probably more): the solution that I've decided on is to support artibitary output formats through the [`--output=template`](https://vale.sh/manual/output/) option.

If you're unhappy with the default output options, you can easily create a custom one without having to modify existing options or continually add new flags.

PR #411, opened 2022-02-22
chore: build darwin_arm64 as well

leipert:
In order to ease installs on M1 Macs, this adds darwin_arm64 to
goreleaser.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -3,12 +3,11 @@ builds:
       - CGO_ENABLED=0
     main: ./cmd/vale/main.go
     binary: vale
-    goos:
-      - windows
-      - darwin
-      - linux
-    goarch:
-      - amd64
+    targets:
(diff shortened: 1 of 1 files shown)


leipert:
👋 This would make our lives at GitLab easier because we could use the procompiled versions on developer machines.

jdkato:
Thanks!

ChrisChinchilla:
Ah fantastic I didn't even think/realise there wasn't a native version. @jdkato does the home-brew formula need updating to add this as an option? I think from what I've seen elsewhere it might do. If that's the case, feel free to assign to me.

jdkato:
The formula is built from source already, so it shouldn't need to be changed.

ChrisChinchilla:
@jdkato Ah ha, OK, so this PR is just for downloaded binaries?

Issue #410, opened 2022-02-18
Error matching when characters like `®` are present in the text


TibsAtWork:
## Summary
Characters like `®` or `™` (U+00AE and U+2122) seem to cause match offset calculations to go wrong.

For instance:
```
$ vale --output=line 'World Health Organization (WHO) (R) and WHO or WHO'
```
but
```
$ vale --output=line 'World Health Organization (WHO) ® and WHO or WHO'
stdin.txt:1:28:Test.WHO_example:'WHO' has no definition
stdin.txt:1:39:Test.WHO_example:'WHO' has no definition
```

## Details

I'm working on macOS 12.1 (Monterey) on an M1 machine.

I've tested against vale versions 2.11.2, 2.14.0 and a build of v2 (which reports v2.15.1). See the "**Exploring further**" section below for some debugging I did using that last.

With my `.vale.ini` as:
```
StylesPath = "styles"

[*]
BasedOnStyles = Vale, Test
```
and minimal setup
```
$ tree styles/
styles/
└─Test/
  └─WHO_example.yml
```
with `styles/Test/WHO_example.yml` as:
```
extends: conditional
message: "'%s' has no definition"
level: error
scope: text
ignorecase: false

first: '\bWHO\b'
second: '(?:\b[A-Z][a-z]+ )+\((WHO)\)'
```

I get:
```
$ vale --output=line 'World Health Organization (WHO) (R) and WHO or WHO'
```
which is a success (if I don't use `--output=line` it reports no errors/warnings), but
```
$ vale --output=line 'World Health Organization (WHO) ® and WHO or WHO'
stdin.txt:1:28:Test.WHO_example:'WHO' has no definition
stdin.txt:1:39:Test.WHO_example:'WHO' has no definition
```

On my mac, I entered the `®` at the command line using `OPT`-`R`, which matches the suggestion at https://en.wikipedia.org/wiki/Registered_trademark_symbol.

To double check, I've also done the same by putting the texts into files (`simple-who-good.txt` and `simple-who-bad.txt`) and run `vale simple-who-good.txt` and `vale simple-who-bad.txt`, and that gives the same results.

I did do a search of the vale repository to see if there were any references to telling vale that text is unicode, but didn't find anything.

## Exploring further

I cloned https://github.com/errata-ai/vale.git and edited `internal/check/conditional.go` to add `"fmt"` to the `import` block and some diagnostic printing (I'm new to go, so there may have been a better way to do this):
``` go
[...]

PR #413, opened 2022-03-02
fix: allow `conditional` style to handle unicode

TibsAtWork:
For instance, when the characters ® or ™ (U+00AE and U+2122) are present
in the text.

Related to #410


jdkato:
Thanks -- this was an oversight, as the other extension points make use of the following function:

https://github.com/errata-ai/vale/blob/2c523911ed29b98f10e47ba94f1b3ae40c94f664/internal/check/definition.go#L137

Could you replace lines 92-93 with a call to this function?

PR #449, opened 2022-06-21
chore: build linux_arm64

### LINKED ISSUE #448: Provide Linux/arm64 binary for use in Docker on m1 macs
Hi, it looks like the arm64 binary is only provided for darwin, having one for linux would be useful for Docker users on apple silicon to avoid running intel-based containers in emulation.

I guess this might be a trivial change since cross-compiling is trivial in go.

Once this is done it would additionally be great if the [docker image](https://hub.docker.com/r/jdkato/vale/tags) was published for [both platforms, amd64 and arm64](https://docs.docker.com/desktop/multi-arch/).

fwilhe:
This is useful for running in Docker on new macs.

Closes #448

Related to #411

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -7,6 +7,7 @@ builds:
       - darwin_amd64
       - darwin_arm64
       - linux_amd64
+      - linux_arm64
       - windows_amd64
 
 archives:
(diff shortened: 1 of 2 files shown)


jdkato:
Thanks!

fwilhe:
Perfect, thanks for the merge. Any chance we get a release soon where this binary will be part of?

jdkato:
Yes, there will be a release later this week.

fwilhe:
Very nice, thank you :)

PR #475, opened 2022-08-18
feat(output) Add sprig library to template output

### LINKED ISSUE #474: New functions to template file output
Hello,

Using the current `2.20.1` version, I try to generate a [valid junit.xml file](https://www.ibm.com/docs/en/developer-for-zos/15.0.0?topic=formats-junit-xml-format).

With the very basic functions available in the templating (with `--output=mytemplate.xml` flag), it is really pain in the ass to compute some basic metrics.
I miss-read the code and though that vale uses the very common [`sprig` library](http://masterminds.github.io/sprig/).
Since now, I found that it is not the case and it is very hard to compute data for a junit file from [`Data` structure](https://github.com/errata-ai/vale/blob/1cefe0bdf8fffb21fdab4ab5983cec7cc899092d/internal/cli/custom.go#L18-L21).

I suggest to add the sprig library to the template rendering.

What do you think?

holyhope:
Close #474 

Signed-off-by: Pierre PÉRONNET <pierre.peronnet@saagie.com>


jdkato:
sprig is a large dependency and includes a lot of functions that are of no relevance to Vale, so I've [included a subset](https://github.com/errata-ai/vale/blob/v2/internal/cli/sprig.go) that I felt covered the common use cases.

What functions were to looking to use specifically?

holyhope:
I use `dict` functions to reorder data  I build a a hash of `checkName: listOfAlerts` then iterate over them. (see <https://github.com/errata-ai/vale/issues/474#issuecomment-1219482633>).

Honestly I may solve my issue in an other way, but I found that this way was the easiest way to do the trick.

By the way, this external library brings a good documentation that can be referred from `vale` docs. So no need to maintain your own.

PR #488, opened 2022-09-15
Adds Severity level to test data line template.

grettke:
Both the CLI template

CLI

and the JSON template

JSON

include the error `Severity: The error level.`

but the `line` does not 

Line-without-severity

so this adds it to make it consistent with the other two

Line-with-severity

I don't know if this is the built in behavior or this is a template only for testing. 

Is this the built in behavior or only for testing?


jdkato:
Please see #379.

grettke:
> Please see #379.

🤦 Sorry I forgot to search existing tickets and use your time on this 🙇.

Gotcha thank you and have a great one! 🙂

PR #497, opened 2022-10-13
feat: load local packages

### LINKED ISSUE #493: Allow local ZIPs in packages
The recent Packages feature looks great! But I noticed that you can only reference ZIP packages from a URL, right? In our case (due to our sync mechanisms) it'd be useful to be able to reference local ZIP files, though.

Something like:

```
Packages = Microsoft, ./path/to/custom.zip, https://github.com/errata-ai/errata.ai/releases/download/v1.0.0/Test.zip
```

cymruu:
Closes #493

### DIFF
--- a/cmd/vale/sync.go
+++ b/cmd/vale/sync.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/errata-ai/vale/v2/internal/core"
+	"github.com/mholt/archiver"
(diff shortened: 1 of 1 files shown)


jdkato (approved):

jdkato:
Thanks -- I'll work on adding a few tests / updating the docs and then put out a new release.

PR #526, opened 2022-11-24
Add OffTokens, OnTokens to support alternatives to "vale off", "vale on"

matchaxnb:
This allows to disable linters as a whole for larger sections of text.

Typically, when I am quoting, I want to turn off vale and prettier in order to avoid messing with the source elements.

Example usage

```vale.ini
OffTokens = vale off, prettier-ignore-start, linters off
OnTokens = vale on, prettier-ignore-end, linters on
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,8 @@ require (
 	github.com/ulikunitz/xz v0.5.10 // indirect
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
 	github.com/xo/terminfo v0.0.0-20210125001918-ca9a967f8778 // indirect
-	golang.org/x/sys v0.0.0-20220319134239-a9b59b0215f8 // indirect
+	golang.org/x/exp v0.0.0-20221114191408-850992195362 // indirect
+	golang.org/x/sys v0.1.0 // indirect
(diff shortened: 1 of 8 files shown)


jdkato:
This seems like an unusual request. The equivalent would be asking the prettier project to support Vale's syntax, no?

In any case, you can already do what you want (effectively turn Vale off entirely for a block) with [`BlockIgnores`](https://vale.sh/docs/topics/scoping/#non-standard-markup).

matchaxnb:
Heh, still think this has value as it would bring better usability to that feature. Using multi-line regex matching gets real ugly real fast. I think that config option brings a better interface onboard.

Prettier's (and other linters') bread and butter is code, so they usually work on ASTs and allow ignoring "blocks", which I believe after reading Vale's code are not the core mechanism as Vale is stream-oriented.

That `BlockIgnores` regex is 1/ using regex so it has a lot of potential for undesired behavior 2/ usable with some arcane, but not the most usable tool in the shed 3/ inconsistent with the easier "vale on", "vale off" syntax which totally makes sense.

By allowing this configuration to exist, you'd make the life of people simpler. I don't see how this would be a unusual request. Please reconsider.

Cheers.

PR #534, opened 2022-12-06
feat: install package from local directory

### LINKED ISSUE #521: Enable non-zipped local packages
I was going through the docs earlier when I noticed that we can [reference local packages as `.zip` files][01] and see that was implemented as part of #493.

For our use case, we're working on a project that intends to publish a Vale package but also consume it in development. We could leverage the new functionality by creating the zip locally and referencing it, but that means there's an additional step every time someone clones the repository or updates a style during development.

If we could _also_ reference a folder containing the styles and/or `.vale.ini` file, that would allow us to tighten up our workflow, requiring only sync instead of zip-and-sync. 

[01]: https://vale.sh/docs/topics/packages/#structure-and-hosting

cymruu:
Closes #521

### DIFF
--- a/cmd/vale/sync.go
+++ b/cmd/vale/sync.go
@@ -74,13 +74,18 @@ func readPkg(pkg, path string, idx int) error {
 }
 
 func loadPkg(name, urlOrPath, styles string, index int) error {
-	if _, err := os.Stat(urlOrPath); !os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


cymruu:
@jdkato thank you for reviewing and merging. I have also mentioned the possibility in docs. Here's the PR https://github.com/errata-ai/vale.sh/pull/22

Issue #542, opened 2023-01-04
Unable to download github package


HeyItsGilbert:
- OS: Windows
- Install method: cocolatey
- Vale version: vale version 2.21.0

I created my own vale style package on Github and I can't seem to get it to download. I thought maybe the zip was built wrong and after some back and forth I think it's correct (a folder with the appropriate name and files in it).

You can see the package here: https://github.com/HeyItsGilbert/vale-agentic/releases/tag/v1.0.0

My vale.ini file:
```
StylesPath = .vscode/styles
MinAlertLevel = suggestion
Vocab = Base
Packages = Microsoft, proselint, write-good, alex, Readability, \
https://github.com/HeyItsGilbert/vale-agentic/releases/download/v1.0.0/agentic.zip

[*.md]
BasedOnStyles = Vale, Microsoft, proselint, Readability, agentic
```

Running vale sync in PowerShell 7.2:
```
> vale sync
 SUCCESS  Downloaded package 'Microsoft'
 SUCCESS  Downloaded package 'proselint'
 SUCCESS  Downloaded package 'write-good'
 SUCCESS  Downloaded package 'alex'
 SUCCESS  Downloaded package 'Readability'
Downloading packages [5/6] █████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████                                        83% | 6sopening source file: open https://github.com/HeyItsGilbert/vale-agentic/releases/download/v1.0.0/agentic.zip: The filename, directory name, or volume label syntax is incorrect.
```

I'm assuming this is possible because it's mentioned in https://vale.sh/docs/topics/packages/#structure-and-hosting

HeyItsGilbert:
Confirmed that putting the zip in the directory and just calling the zip files works.  Also confirmed that the issues happens in Windows PowerShell, and cmd shell.

https://github.com/errata-ai/vale/blob/v2/cmd/vale/sync.go#L77
I'm guessing that this is a different error then `!os.IsNotExist(err)`.  Would it be better to just replace `err != nil`? Or possibly add additional cases (such as invalid argument).

I saw the other os.stat errors here https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/os/error.go;l=19;drc=9955a7e9bb40d28502fbb8fd6ef1f2f10e18a519. That being said, I haven't written much go so I may be over simplifying.

PR #543, opened 2023-01-04
fix: allow for more generic errors in loadPkg

HeyItsGilbert:
fix: allow for more generic errors in loadPkg

os.Stat() was receiving an error but the conditional was only looking for a "not exist" error. This change makes it so that we skip if there are ANY errors.

Related to #542


HeyItsGilbert:
I confirmed this by running: ` env GOOS=windows go build` and downloading the artifact from my codespace. Pointed that to my vale.ini with a URL and it succeeded!

jdkato:
Thanks!

PR #558, opened 2023-01-28
feat: Add pre-commit support

AleksaC:
This enables running vale with [pre-commit](https://github.com/pre-commit/pre-commit).

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+- id: vale
+  name: vale
+  description: Run vale on your text
+  entry: vale
+  language: golang
+  types: [text]


JacobCoffee:
As requested in discussion #556

jdkato:
Thanks!

jarmos-san:
Thanks for the PR but it appears there's some issue with the `.pre-commit-hooks.yaml` file. I'm stumbling across this error message after running the `pre-commit install --install-hooks` command;

```console
[INFO] Initialising environment for https://github.com/errata-ai/vale
An error has occurred: InvalidManifestError:
=====> ~/.cache/pre-commit/repokv40qx5e/.pre-commit-hooks.yaml is not a file.
```
Perhaps running the `pre-commit validate-manifest` command will suggest something about the file to be fixed?

AleksaC:
Which revision of the repo are you using? I think it doesn't work on revisions that don't have the `.pre-commit-hooks.yaml` file. Currently the only one that would work is `16d3a7f`. This works for me and should work in general:

```yaml
repos:
  - repo: https://github.com/errata-ai/vale
    rev: 16d3a7f
    hooks:
      - id: vale
```

jarmos-san:
Got it, I was using the latest tagged version which obviously doesn't have the `.pre-commit-hooks.yml` file!.

PR #598, opened 2023-03-28
(GH-597) Find config in home folder as last resort

### LINKED ISSUE #597: Vale unable to find default config in home folder
On both Windows and in the Windows Subsystem for Linux (WSL), Vale doesn't seem to be able to find the `.vale.ini` configuration file in my home directory. Based on the [documentation][01] and the [code comments][02], I expected Vale to use `~/.vale.ini` as the default configuration when no configuration file exists closer to documents I want to lint.

Looking through the code, I suspect that it never finds `~/.vale.ini` because it's only recursing back up the path from the current working directory - if I'm working on documents anywhere else (very common on Windows), Vale never checks the home folder for the current user.

https://github.com/errata-ai/vale/blob/11fdc85dd4f7928fb35bf5312cbc783c327122fc/internal/core/ini.go#L215-L243

> Edit:
> 
> I was just able to confirm that if I copy my project into a subfolder arbitrarily nested under my home folder that Vale is able to find the existing `~/.vale.ini` file. So this issue is only (but always) present when the working directory doesn't have the home folder as an ancestor.

## Install Details

- OS: Windows, WSL (Ubuntu)
- Install method: Direct download
- Vale version: `vale version 2.24.0`

[01]: https://vale.sh/docs/topics/config/#search-process
[02]: https://github.com/errata-ai/vale/blob/11fdc85dd4f7928fb35bf5312cbc783c327122fc/internal/core/ini.go#L215-L217

michaeltlombardi:
Prior to this change, the [documentation][01] noted that when no closer configuration can be found, Vale will default to using the configuration file in the user's home directory. However, the implementation for the `loadConfig()` function only recursed through the current working directory's ancestors.

This meant that if the working directory didn't have the user's home directory as an ancestor, no configuration would be found and the user would get an error even when they had defined a configuration file in their home directory.

This change:

- Updates the `loadConfig()` function to explicitly search the user's home directory if no configuration file is discovered in an ancestor folder of the current working directory.
- Maintains all other prior functionality.
- Resolves #597.

[01]: https://vale.sh/docs/topics/config/#search-process

### DIFF
--- a/internal/core/ini.go
+++ b/internal/core/ini.go
@@ -212,9 +212,10 @@ func loadINI(cfg *Config, dry bool) error {
 	return processConfig(uCfg, cfg, sources, dry)
 }
 
-// loadConfig loads the .vale file. It checks the current directory up to the
-// user's home directory, stopping on the first occurrence of a .vale or _vale
-// file.
+// loadConfig loads the .vale file. It checks the ancestors of the current
+// directory, stopping on the first occurrence of a .vale or _vale file. If
+// no ancestor of the current directory has a configuration file, it checks
+// the user's home directory for a configuration file.
 func loadConfig(names []string) (string, error) {
 	var parent string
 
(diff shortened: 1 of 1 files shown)


jdkato (approved):

jdkato:
Thanks!

theletterf:
This seems to have broken compatibility with Vale Server. Could it be?

PR #682, opened 2023-09-13
Feat: Add Optional Custom Message to Script Match

p0fi:
# 📖 Description
This PR implements the feature suggested in https://github.com/errata-ai/vale/discussions/679

The `matches` object populated by a Tengolang script in a script based rule can now also have an optional `message` per match. This way, a single script-based rule can find several similar issues while still providing a precise error message to the user on a per-match basis.


jdkato (approved):

jdkato:
Thanks! I'll add some tests and then put out a new release.

PR #731, opened 2023-11-18
Julia Support

### LINKED ISSUE #729: Add Julia support
### Check for existing issues

- [X] Completed

### Describe the feature

It seems that in this list

https://vale.sh/docs/topics/scoping/#code-1

Julia is not yet included.

Single comments are started with `# (comment)`
Multiline comments are between `#= ... =#`

and (cf https://github.com/errata-ai/vale/discussions/728) if doc strings (or strings in general?) could also be supported as scope that would be awesome!

kellertuer:
This PR aims to add Julia support (resolves #729).

I would like to also try to resolve #728 if possible. I added a test file, but I am unsure how to accomplish the string part.
I am also not sure whether my code is correct, since I have never programmed in Go before nor did I have the time to install that, so I just tried to mimic what was there.

For the doc strings, see https://docs.julialang.org/en/v1/manual/documentation/ , maybe it is really just enough to consider strings and multiline strings as text.

The only caveats there are

* `"$(a) and $a"` is a string where the current value of `a` is interpolated twice (placed in there
* inline math in Julia flavoured Markdown is done with two backticks like ``` raw"``\frac{1}{2}``" ```, where the keyword `raw` is just used so that I do not have to escape the backslash.

(but maybe the two backpacks can be added to some ignore in the `.vale.ini` like other Markdown inline math ones as well)

Thanks for writing such a nice package, I already use it on a Julia Package (Markdown) Documentation I am writing.

### DIFF
--- a/internal/core/format.go
+++ b/internal/core/format.go
@@ -66,6 +66,11 @@ var CommentsByNormedExt = map[string]map[string]string{
 		"blockStart": `(^=begin)`,
 		"blockEnd":   `(^=end)`,
 	},
+	".jl": { // Julia
+		"inline":     `(#.+)`,
+		"blockStart": `(#=)`,
+		"blockEnd":   `(=#)`,
+	}
 }
 
 // FormatByExtension associates a file extension with its "normed" extension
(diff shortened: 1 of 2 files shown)


kellertuer:
I read a bit about regexpses and looked at the Python example. The current code now tries to include (doc) strings – I added a bit of comments, since for the beginning block I am not sure that works as intended. I put all cases now also in the example file, which. still has to be added to the tests – I have not yet understood how to do that, especially not which output to expect, since I was not yet able to set vale up locally from source..

jdkato:
Thanks -- setting up an example file is a big help. I can add the tests and then this should be ready for the next release.

kellertuer:
Great, if you need help with the file or understanding something on how Julia is formatted usually, let me know.

jdkato:
This should be working in the latest release. Let me know if you run into any issues.

PR #748, opened 2024-01-03
feat: add comment syntax for switching styles

ak-1:
This adds some comment syntax for controlling styles.

Individual styles can be switched on and off:
```
<!-- vale StyleName1 = YES -->
<!-- vale StyleName2 = NO -->
```
For a rule to be effective both the rule and the style has to be enabled.

`on` and `off` is accepted in addition to `YES` and `NO`
```
<!-- vale StyleName1 = on -->
<!-- vale StyleName2 = off -->
```

Styles can be set (enabling them and switching off any other styles):
```
<!-- vale style = StyleName1 -->
<!-- vale styles = StyleName1, StyleName2 -->
```

### DIFF
--- a/internal/core/file.go
+++ b/internal/core/file.go
@@ -16,7 +16,9 @@ import (
 	"github.com/errata-ai/vale/v2/internal/nlp"
 )
 
-var commentControlRE = regexp.MustCompile(`^vale (.+\..+) = (YES|NO)$`)
+var commentControlRE = regexp.MustCompile(`^vale (.+\..+|[^.]+) = (YES|NO|on|off)$`)
+
+var commentStyleRE = regexp.MustCompile(`^vale styles? = (.*)$`)
 
 // A File represents a linted text file.
 type File struct {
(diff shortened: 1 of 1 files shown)


jdkato:
Sorry for the delay here. The changes look good.

Once I add tests and update the docs, I'll put out a new release.

jdkato:
This you test this locally? 

I tried to write tests for it and doesn't seem to work, and I really don't have the time to re-implement it myself.

Some thoughts:

- I think we should skip the "Styles can be set" functionality. At best, it will only work for styles that have *already* been enabled through the `.vale.ini`. In other words, it's the same thing as disabling certain styles.
- The options would then be: (1) disable a rule, (2) disable a style, or (3) disable Vale.

If you'd like to add tests yourself, the existing ones are here: https://github.com/errata-ai/vale/tree/v3/testdata/fixtures/comments

ak-1:
Thanks for looking into this. I don't know why this did not work for you though. I am using these changes for a while now without any issues. I will post a new PR with tests included.

Regarding  "Styles can be set" functionality: Maybe it depends on the use case, but I find the "vale style = ..." syntax much nicer in practice than enabling/disabling individual styles. With "set style" it is really obvious what styles are in effect for each section. Using enable/disable is more complicated/error prone because you have to track manually what styles are in effect at any point.

For example, when writing fiction you could have different chapters written in different tenses and point of view etc. Then you can create a style for each POV, setting, tense, epoch, person, etc. and list at the top of each chapter what styles apply. This is not easy to replicate with just enable/disable.

It is true that you still have to list all the styles in .vale.ini, but that is also the case for enable/disable.

Let me know what you think. If you don't find that convincing I can split up the individual features.

PR #772, opened 2024-02-17
feat: add comment syntax for switching styles

ak-1:
This is a follow-up to #748 with tests included.

Individual styles can be switched on and off:
```
<!-- vale StyleName1 = YES -->
<!-- vale StyleName2 = NO -->
```
For a rule to be effective both the rule and the style has to be enabled.

`on` and `off` is accepted in addition to `YES` and `NO`
```
<!-- vale StyleName1 = on -->
<!-- vale StyleName2 = off -->
```

Styles can be set (enabling them and switching off any other styles):
```
<!-- vale style = StyleName1 -->
<!-- vale styles = StyleName1, StyleName2 -->
```

### DIFF
--- a/internal/core/file.go
+++ b/internal/core/file.go
@@ -15,7 +15,9 @@ import (
 	"github.com/errata-ai/vale/v3/internal/nlp"
 )
 
-var commentControlRE = regexp.MustCompile(`^vale (.+\..+) = (YES|NO)$`)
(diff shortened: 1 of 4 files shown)


jdkato:
Thanks for your work here!

ccoVeille:
Thanks it's what I needed for some files

PR #777, opened 2024-02-19
feat: proto support

mfranberg:
Hi!

This PR adds support for .proto files, using the same comment setup as the c-formatter.


jdkato:
Thanks!

PR #842, opened 2024-05-30
Add xsltproc and dita to prerequisites

biinari:
In order to run the cucumber tests, `xsltproc` and `dita` are both required to be available in the `$PATH`.


jdkato:
Thanks!

PR #844, opened 2024-05-31
feat: add comment syntax for ignoring individual matches

ItsVeryWindy:
This allows you to turn off rules for specific matches. I was mainly thinking of situations where you may want to check if a word is being used correctly in a paragraph, and then receive no more notifications about it, but still allow other words to be picked up. 

This is the syntax I've gone for, where the array can contain one or more matches to disable:

```
<!-- vale vale.Redundancy["ACT test"] = NO -->
<!-- vale vale.Redundancy["ACT test"] = YES -->
```

I haven't implemented all the tests yet, looking to get feedback if this is a proper way of solving the problem.


jdkato:
Thanks for your work here. I'll add the remaining tests and try to get this into the next release.

PR #861, opened 2024-07-04
Add support for the dart language

rafaeltab:
Hey, I wanted to have support for the dart language, so I added it. Please let me know if I missed something.

I added `inBlock` to the `comments.go` file because when using a `/**` comment people often add a `*` prefix as such:
```dart
/*
 * Hey
 * How
 * Are
 * You
 */
 ```
 
 This is removed by the `inBlock` regex.
 
In the `format.go` file I did not add this, as it seems that each regex includes the comment in the selection here, so there is no need to remove anything, if I am mistaken please let me know.

> [!IMPORTANT]  
> Dart actually uses markdown inside their comments, I was not sure how to make it recognize it as such. Comments are only interpreted as normal text.

### DIFF
--- a/internal/core/format.go
+++ b/internal/core/format.go
@@ -57,6 +57,11 @@ var CommentsByNormedExt = map[string]map[string]string{
 		"blockStart": `(^=begin)`,
 		"blockEnd":   `(^=end)`,
 	},
+	".dart": {
+		"inline":     `(/{2,3}.+)`,
+		"blockStart": `(\/\*{1,2}.*)`,
+		"blockEnd":   `(.*\*{1,2}\/)`,
+	},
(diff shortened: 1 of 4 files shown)


jdkato:
I've moved away from regex-based comment extraction in favor of using Tree-sitter. You can see examples in the `internal/lint/code` package.

> Dart actually uses markdown inside their comments, I was not sure how to make it recognize it as such. Comments are only interpreted as normal text.

Vale can handle this for all supporting languages.

rafaeltab:
Oh awesome! I actually looked at adding dart support before, and found a branch with treesitter, but then I couldn't find it this time. I'll get right on implementing this.

PR #900, opened 2024-09-23
feat: Allow users to override comment delimiters

### LINKED ISSUE #762: Support custom comment patterns (e.g., MDX)
### Check for existing issues

- [X] Completed

### Describe the feature

Vale version: v3.0.6 (at `55f450fe90d954b9c2019c36560472603e925cc3`)

## Summary

When using Vale with MDX files, I've mostly found success by setting `mdx = md` in the `[formats]` field of `.vale.ini`. One incompatibility between MDX and MD will make it difficult to implement Vale in my organization, though: the fact that Vale ignores MDX comments, which use `{/* This syntax */}`. Since Vale enables you to temporarily switch a rule on and off using comments, ignoring MDX comments means that docs authors can't configure Vale to ignore specific document elements. This will make it cumbersome for docs authors who run into edge cases. 

## Testing

Let's say we want to lint this MDX document:

```
 # Hello

This is a paragraph.

{/* SHOULD BE A COMMENT */}

End of doc.
```

The `.vale.ini` file includes this setting:

```ini
[formats]
mdx = md
```

I ran `vale` with `dlv` to lint the example file and set a breakpoint when we use Goldmark to convert the Markdown doc to HTML:

https://github.com/errata-ai/vale/blob/55f450fe90d954b9c2019c36560472603e925cc3/internal/lint/md.go#L41-L43

After this, the resulting HTML fragment looks like this:

```html
"<h1>Hello</h1>\n<p>This is a paragraph.</p>\n<p>{/* SHOULD BE A COMMENT */}</p>\n<p>End of doc.</p>\n"
```
The MDX comment is in an HTML paragraph tag.

## Possible solution

Since MDX does not allow for standard Markdown comments (`<!-- -->`), supporting MDX requires allowing custom comment delimiters.

We could add a config field called `CommentPattern` or similar that takes a regular expression. In `lintMarkdown`, before applying the Goldmark conversion, we could replace instances of the `CommentPattern` regex with a standard HTML comment tag. When we apply the conversion, comments would be parsed as expected.

I'm happy to take this on if you don't have the bandwidth. Thanks!

ptgott:
Allow users to specify the comment delimiters they are using in their documentation. vale replaces these with HTML comment tags before linting, making it possible to control style rules for specific passages of prose in file formats that use non-HTML comment syntax. This is critical for controlling style rules within a page in MDX, and potentially other formats as well.

This example configures the `CommentDelimiters` field for `*.md` files, indicating that `{/*` and `*/}` are the custom comment delimiters:

```ini
[*.md]
CommentDelimiters = "{/*,*/}"
```

Internally, custom delimiters are represented as a `[2]string`, and it is only possible to configure one set of custom comment delimiters for a given format block.

More specific changes:
- Refactor `applyPatterns`. Remove the method receiver and take only the necessary fields of `*core.Config` as parameters. This makes it easier to test `applyPatterns` without mocking an entire `*core.Config`. Also extract functions for `applyInlinePatterns` and `applyBlockPatterns` so we can use fewer arguments in a single function.
- Add `applyCommentPatterns`, which works similarly to `applyInlinePatterns` and `applyBlockPatterns`, but for substituting comments.
- For tests, import `https//github.com/stretchr/testify/assert`, a popular testing library, to get richer test output.

Closes #762

### DIFF
--- a/internal/lint/html_test.go
+++ b/internal/lint/html_test.go
@@ -1,0 +1,167 @@
+package lint
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/errata-ai/vale/v3/internal/core"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_applyPatterns(t *testing.T) {
+	cases := []struct {
+		description string
+		conf        core.Config
+		exts        extensionConfig
+		content     string
+		expected    string
+	}{
+		{
+			description: "MDX comment in markdown, custom comment delimiter",
+			conf: core.Config{
+				CommentDelimiters: map[string][2]string{
+					".md": [2]string{"{/*", "*/}"},
+				},
+			},
+			exts: extensionConfig{".md", ".md"},
+			content: `---
+title: Example page
+description: Example page
+---
+
+This is the intro pagragraph.
+
+{/* This is a comment */}
+`,
+			expected: strings.ReplaceAll(`
+@@@
+title: Example page
+description: Example page
+@@@
+
+
+This is the intro pagragraph.
+
+<!-- This is a comment -->
+`, "@", "`"),
+		},
+		{
+			description: "MDX comment in markdown, no custom comment delimiter",
+			conf:        core.Config{},
+			exts:        extensionConfig{".md", ".md"},
+			content: `---
+title: Example page
+description: Example page
+---
+
+This is the intro pagragraph.
+
+{/* This is a comment */}
+`,
+			expected: strings.ReplaceAll(`
+@@@
+title: Example page
+description: Example page
+@@@
+
+
+This is the intro pagragraph.
+
+{/* This is a comment */}
+`, "@", "`"),
+		},
+		{
+			description: "multiline MDX comment in markdown, custom comment delimiter",
+			conf: core.Config{
+				CommentDelimiters: map[string][2]string{
+					".md": [2]string{"{/*", "*/}"},
+				},
+			},
+			exts: extensionConfig{".md", ".md"},
+			content: `---
+title: Example page
+description: Example page
+---
+
+This is the intro pagragraph.
+
+{/* 
+This is a comment 
+*/}
+`,
+			expected: strings.ReplaceAll(`
+@@@
+title: Example page
+description: Example page
+@@@
+
+
+This is the intro pagragraph.
+
+<!-- 
+This is a comment 
+-->
+`, "@", "`"),
+		},
+		{
+			description: "token ignore in cc file",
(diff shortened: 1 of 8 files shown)


ptgott:
Hello @jdkato, it looks like a number of cucumber tests are failing in CI. When I set up a development environment per the [instructions](https://github.com/errata-ai/vale/blob/v3/.github/CONTRIBUTING.md#setting-up-a-development-environment) and run `make test` on `origin/v3`, cucumber tests also fail. Is this a known issue? Is there a way to verify which failures were introduced by my changes? Thanks!

jdkato:
The tests pass on the v3 branch on GitHub and for me locally.

So my guess is that the failures you're seeing locally on the v3 branch are related to your environment, but I would need to know which cases are failing to say more.

The failures here are related to your changes, but I haven't looked closely enough to say exactly why.

My advice would be to try running Vale in the relevant test directories, starting with `testdata/fixtures/patterns`.

ptgott:
@jdkato Thanks for your help! I realized why my changes were introducing failures and pushed a fix with a new unit test.

jdkato:
The changes look good. Are you ready for this to be merged?

internal/lint/html_test.go
```diff
@@ -0,0 +1,167 @@
...
+This is a comment 
+-->
+`, "@", "`"),
+		},
+		{
+			description: "token ignore in cc file",
+			content:     "Call \\c func to start the process.",
+			conf: core.Config{
+				TokenIgnores: map[string][]string{
+					"*.cc": []string{"(\\\\c \\w+)"},
```
ccoVeille: When I see such a thing, I always wonder if using string literals (backtick notation) won't help

Something like

```suggestion
					"*.cc": []string{`(\\c \w+)`},
```

Or this


```suggestion
					"*.cc": []string{"(\c \w+)"},
```

Because I'm unsure why you could need a double escaped c, but a simple escaped w.

Unless you are looking for the characters `\c` followed by a space then one or many alphanumeric characters.

Please don't focus on my misunderstanding about what you are looking for, and consider the backtick notation anyway

ptgott: Thanks! Fixed in [7ed02e6f92b311dff574f2eb57addc78f0b47eec](https://github.com/errata-ai/vale/pull/900/commits/7ed02e6f92b311dff574f2eb57addc78f0b47eec) . I think I'd copied a value from `dlv` to write the test and didn't consider cleaning it up.

ptgott:
> The changes look good. Are you ready for this to be merged?

Yes, thanks for taking a look and helping me through the process!

Issue #929, opened 2024-12-06
Zig language support


LDprg:
### Check for existing issues

- [X] Completed

### Describe the feature

It would be great to have zig language support for vale

PR #930, opened 2024-12-06
feat: zig support

LDprg:
This is related to #929. I tried adding my own zig support, not sure if I did everything right.

I just want to note that zig does only support line comments (`//` or `///` or `//!`).

### DIFF
--- a/internal/core/format.go
+++ b/internal/core/format.go
@@ -54,7 +54,7 @@ var CommentsByNormedExt = map[string]map[string]string{
 		"blockEnd":   `(.*-\})`,
 	},
 	".zig": {
-		"inline":     `(//.+)`,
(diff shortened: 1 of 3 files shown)


jdkato:
I'd like to add support for new code languages through tree-sitter, rather than the now-deprecated method of regex.

Vale uses [github.com/smacker/go-tree-sitter](https://github.com/smacker/go-tree-sitter), which doesn't include Zig by default.

You can either contribute to that repo or create your own Go package for Zig.

I did something similar for Julia: [package](https://github.com/jdkato/go-tree-sitter-julia), [source](https://github.com/errata-ai/vale/blob/v3/internal/lint/code/jl.go).

PR #935, opened 2024-12-15
fix: check `qErr != nil` but return a nil value error `err`

alingse:
the p.Kill() return an error, but return a nil

I create a linter to detect code that returns a non-relevant nilness error bug. I checked the top 1000 GitHub Go repositories and found this, all result listed in https://github.com/alingse/sundrylint/issues/4


ccoVeille:
Hi @alingse 

Thank you.

I'm glad you did, I'm also using vale

PR #938, opened 2024-12-17
chore: update vulnerable go dependencies

cheungaryk:
Hello, this is related to #936, Richard (that PR's submitter) and I work for the same organization.

We are trying to get vale to be approved for use in our organization. The security review found multiple vulnerabilities that need to be patched before approved. So I created this PR and also #937 .

This PR patched the vulns related to go. Most of them are pretty straight forward patch/minor upgrades, here are the more notable ones:

- [antonmedv/expr](https://pkg.go.dev/github.com/antonmedv/expr) -> [expr-lang/expr](https://pkg.go.dev/github.com/expr-lang/expr). The author renamed it, it's the same library
- [mholt/archiver/v3](https://github.com/mholt/archiver) -> archive/zip (go native). Per mholt the author, archiver is deprecated and he has moved the work to a newer library named [archives](https://github.com/mholt/archives), but in this PR I am using archive/zip instead for simplicity (and also archives doesn't pass our security scan 😓 ). Comes with slight code refactoring. **Note that in order to resolve the "decompression bomb" lint warning, I imposed a 10GB file size limit.** Please let me know if the file size is too large (or too small)
- [pterm](https://pkg.go.dev/github.com/pterm/pterm) -> downgrade from 0.12.76 to 0.12.40. The dependencies are different in both versions, but our security scan rejects the indirects above 0.12.40
- [godirwalk](https://github.com/karrick/godirwalk) -> path/filepath (go native). Only the latest version, 1.7.0, is accepted by our security scan, but 1.7.0 reportedly has a bug (https://github.com/karrick/godirwalk/issues/70) that is not being fixed or responded to. Also the author wrote this on the readme: "Depending on your specific circumstances, [you might no longer need a library for file walking in Go](https://engineering.kablamo.com.au/posts/2021/quick-comparison-between-go-file-walk-implementations).", linking to an article that seems to imply the go native filepath is now faster than godirwalk

All CI/CD test cases have passed, so hopefully this PR is acceptable.

Thank you!


jdkato:
Thanks for your work here.

In the future, I'd ask to make individual PRs for dependency changes/removals that require code changes so it's easier to review.

For example, the `godirwalk` removal is technically a breaking change since it doesn't handle symlinks the same way. However, I don't yet have tests that capture this, so I'll work on this before the next release.

cheungaryk:
Many thanks @jdkato. If you have a moment, would you mind manually triggering a release? The CI/CD pipeline failed due to 502 error.

PR #944, opened 2025-01-05
fix: allow optional language in code block

ArnaudD-FR:
According to [sphinx documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block) the language name of a code block is optional. By the way some directive options may have no argument.

This PR fix update the regular expression `reCodeBlock` with the following changes:

 * spaces patterns before language name and option values are expected 0 or many times
 * language name pattern is expected 0 or 1 time
 * directive dots are only dots and not any character
 * rst argument of directive option can be optional


jdkato:
Thanks!

PR #993, opened 2025-04-07
feat: add replace action to default Vale.Terms rule

wbh1:
This is useful for ensuring that Vocab terms get populated as quick fix/code action suggestions by the Vale Language Server.

### DIFF
--- a/internal/check/definition.go
+++ b/internal/check/definition.go
@@ -79,6 +79,9 @@ var defaultRules = map[string]map[string]interface{}{
 		"swap":       map[string]string{},
 		"vocab":      false,
 		"path":       "internal",
+		"action": core.Action{
+			Name: "replace",
+		},
 	},
 	"Repetition": {
 		"extends":    "repetition",


jdkato:
The reason why this wasn't added before is because vocabularies can (and often do) contain regex instead of words or phrases. In these cases, the suggestions won't make sense.

ccoVeille:
Good point, indeed.

This is the default rule, right?

Would it be possible to add the replace action on some condition only? Like the words is only Unicode letters (naive idea for detection) for example. But then, I assume it wouldn't be there, but where the code detects the words is invalid.
So maybe the feature idea could be kept by moving it somewhere else, no?

Another question, you said the vocabulary contains (often) regexp so it cannot be done with the way the author of the PR did. But I see the message field of the rule saying "use %s instead of %s" so I assume the word provided for the first %s is a valid word, that people would expect to see used with a "quick action".

jdkato:
> Would it be possible to add the replace action on some condition only? Like the words is only Unicode letters (naive idea for detection) for example. But then, I assume it wouldn't be there, but where the code detects the words is invalid.
So maybe the feature idea could be kept by moving it somewhere else, no?

Yes -- honestly, I'd like to move vocabularies to YAML files as well. This would make them much more robust in many ways (comment handling, feature additions, conditional entries, etc).

ccoVeille:
I would love that, yes

jdkato:
I'm going to try to get support for YAML-formatted vocabularies into the next release.

PR #1016, opened 2025-06-16
feat: add support for Nix files

thoughtpolice:
I added a test for this that I think is correct, but I haven't been able to run it yet because I can't quite get cucumber properly installed (`make setup` fails for me...) Otherwise I think this is correct? Let me know if anything needs to change, thanks!


thoughtpolice:
(I largely just based this on 6ef0894f7bca5342762e0c34ac89ea27da67)

thoughtpolice:
Oh, I see new things should use tree-sitter. I'll look into that.

PR #1066, opened 2026-01-16
Fix handling of xslt path and possible args pollution for xslt calls

moritzfl:
This fixes the handling of the path to xslt files along with another bug that I found along the way where vale would incorrectly compose command line arguments for xslt.


jdkato:
Thanks!

PR #1067, opened 2026-01-16
Fix #1046 - handle xsd as xml

### LINKED ISSUE #1046: It should be possible to handle XSD files as XML
### Check for existing issues

- [x] Completed

### Describe the feature

We create XSD schema files with documentation within them.

It took me a while to notice why they were not processed by vale - they are simply not considered xml and throwing a "Transform" at them does not do anything.

This is what I tried - renaming the xsd files to xml produced the desired result but that will not work for our software repo. 

All xsd files can only be processed as raw text files.
```
[Configuration/*.{xml,xsd}]
Transform = xslt/xml-multilang-html.xsl
BasedOnStyles = English, German
```

It would be nice if we could get the possibility to check xsd files as well.

moritzfl:
This activates handling of xsd files as xml.


jdkato:
Thanks!

PR #1071, opened 2026-02-04
feat: add ignore field to View scopes for rule filtering

jyaguem:
Allow specifying rules to skip when linting specific YAML keys in Views. This enables ignoring certain rules (e.g., spelling) for metadata fields while still applying them to content fields like descriptions.

Example usage in a View file:
```
  scopes:
    - name: metadata 
      expr: name 
      ignore: 
        - Vale.Spelling
```


jdkato:
This should be done in the other direction: scopes don't choose their rules, rules choose their scopes. See https://vale.sh/docs/scopes#selectors

PR #1072, opened 2026-02-06
feat(lint): apply defaultIgnoreDirectories to nested directories

secustor:
### Context

Currently vale will still lint directories like `node_modules` or `.git` folders, if the are not at the root directory. 

Example cases could be: 
- JS monorepo `packages/foo/node_modules` 
- Git Worktree setup `worktree-a/.git`

### Implementation

This PR changes ignore behaviour for `defaultIgnoreDirectories` from being the full path to being the name of lowest directory name.


jdkato:
Thanks!

PR #1076, opened 2026-02-09
feat: set path for stdin cli

moritzfl:
This concerns command line use and the use of the command line for IDE plugins that use the command line to interact with vale.

Setting the file extension when checking file content through command line, it was not possible to lint with path based vale configurations. For example any rule set to check `*/documentation/en/*.md` would not hit any markdown files because only the extension was passed through the cli.

This pull request adds the "path" parameter to the cli. This allows better integration into IDE plugins that may send the current unsaved file content to vale for a check but then also need to pass a path in order for rules to match properly.

The underlying issue has also been discussed to some extent here (Vale Plugin for IntelliJ): https://gitlab.com/pablomxnl/vale-cli-plugin/-/merge_requests/49


jdkato:
Thanks!

Issue #869, opened 2024-07-10
Regex caret (^) not working properly for scope: raw when there are over 999 characters in a Markdown file.


michael-nok:
### Check for existing issues

- [X] Completed

### Environment

- Windows 10
- Direct download of Windows executable
- Vale 2.29.1 or later

### Describe the bug / provide steps to reproduce it

The changes implemented by this commit causes the problem to exist: [f769fcd](https://github.com/errata-ai/vale/commit/f769fcd3aeb7606a62cf1a0da09fb96c8dfc6e31)

```
	// NOTE: If the `ctx` document is large (as could be the case with
	// `scope: raw`) this is *slow*. Thus, the cap at 1k.
	//
	// TODO: Actually fix this.
```

I have a rule that looks for incorrectly indented content. It uses the following token:

```
extends: existence
message: 'Content must be indented using 4x spaces each time. "%s"'
level: error
nonword: true
scope: raw
tokens:
  - '^[ ]{1,3}\`'
```

When the Markdown file contains 999 characters or more (i.e. `ctx` > 1000), the `^` part of the token stops using the start of the line properly and invents (hallucinates) new starting positions.

Attached are sample files that exactly show the spillover in the logic:
[vale.zip](https://github.com/user-attachments/files/16169264/vale.zip)

Consequently, text with four spaces before the first \` is flagged as incorrect, and the starting position for the ^ is column 1.

![image](https://github.com/errata-ai/vale/assets/67281799/b1445dc8-5690-4f75-aaee-7a23b8b604b3)

Using the vale.exe from release 2.29.0 does not have this problem.

twitchard:
I ran into this too.

My reproduction
```
cd $(mktemp -d)                                                             
cat <<EOF > .vale.ini
StylesPath = ./.vale/styles
MinAlertLevel = suggestion
[*.txt]
BasedOnStyles = MyStyle
EOF
mkdir -p ./.vale/styles/MyStyle
cat <<EOF > .vale/styles/MyStyle/MyRule.yml
extends: script
message: "Error %s"
scope: raw
script: |
  matches := [{begin: 0, end: 1}, {begin: 997, end: 998}]
EOF
(
  echo a
  echo b
  for x in `seq 993`; do
    echo -n "x";
  done
  echo b
) >> test-999-bytes.txt
cp test-999-bytes.txt test-1001-bytes.txt
echo x >> test-1001-bytes.txt
vale test-999-bytes.txt
vale test-1001-bytes.txt
[...]

Issue #272, opened 2020-11-16
Exclusions don't seem to work properly in scope: raw


Ravlen:
Another minor bug, but it seems like exclusions don't work as expected for rules using `scope: raw`. I can add the standard exclusions for all of our rules that *don't* use `scope: raw`, but none of the rules that DO use it can't be excluded properly.

See how the first rule gives suggestions, then within the exclusion block it does not. Then the second rule (which uses `scope: raw` does not exclude the alert:

<img width="588" alt="Screen Shot 2020-11-16 at 16 00 34" src="https://user-images.githubusercontent.com/36979740/99221980-fb00cf00-2824-11eb-81b5-90e9ee55259b.png">

The weird thing is that if you forget to add the second line (to re-enable the rule), it then excludes the rule from the entire file, including lines BEFORE the exclusion. Non-`scope: raw` rules work as expected, only excluding everything after the exclusion line:

<img width="875" alt="Screen Shot 2020-11-16 at 16 03 54" src="https://user-images.githubusercontent.com/36979740/99222232-78c4da80-2825-11eb-8ae5-d99b79b5d449.png">

cc @jdkato

jdkato:
This is the same issue as #194: any HTML-related scoping / configuration features aren't compatible with `scope: raw` (this includes HTML comments, `BlockIgnores`, `TokenIgnores`, `IgnoredClasses`, and `SkippedScopes`).

> The weird thing is that if you forget to add the second line (to re-enable the rule), it then excludes the rule from the entire file, including lines BEFORE the exclusion. 

It's not clear to me what's happening here since this shouldn't be related to ignore comments (as mentioned above). Is it possible that the comment placement is interfering with the rule's regex?

Ravlen:
@jdkato Ahh, when it started to work partially (when I added the first line), I thought you might have implemented it and I just missed it in a minor release or something.

It's definitely not being caught by the regex, I think. I changed the regex to be just `Internal`, and the behavior is the same. Add the exclusion start, and it's turned off for the whole page. Add the exclusion end, and it's as if there is no exclusion at all.

<img width="701" alt="Screen Shot 2020-11-18 at 15 57 57" src="https://user-images.githubusercontent.com/36979740/99496041-ed7b4e80-29b6-11eb-860e-319048454b7b.png">

[...]

PR #1081, opened 2026-02-23
fix(core): use byte offsets for position reporting in raw-scoped script rules

### LINKED ISSUE #1083: scope: raw script rules report wrong position when matched text appears multiple times
### Check for existing issues

- [x] Completed

### Environment

- **Vale version**: 3.13.1 (also reproduces on 3.12.0)
- **Operating system**: Linux
- **Installation method**: Source / Docker

When a Tengo script rule uses `scope: raw`, Vale does not use the `begin` and `end` byte offsets returned by the script to calculate the alert's line and column. Instead, Vale extracts the matched text (`scope[begin:end]`) and performs a **text search** in the parsed document to determine the position. If the matched text appears multiple times in the document, Vale always reports the position of the **first** occurrence, regardless of which occurrence the script intended to flag.

### 1. Create a script rule

`styles/Example/FindTODO.yml`:

```yaml
extends: script
message: "Found a TODO."
scope: raw
level: warning
script: find-todo.tengo
```

`styles/config/scripts/find-todo.tengo`:

```tengo
text := import("text")
matches := []

// Hardcode a match at byte offset 54, which is the second
// occurrence of "TODO" in the test document (line 3).
matches = append(matches, {
    begin: 54,
    end:   58
})
```

### 2. Create a test document

`test.md` (the word "TODO" appears on lines 1 and 3):

```markdown
# TODO list for the project

This paragraph has a TODO that should be flagged by the script rule.
```

- `TODO` first appears at byte offset 2 (heading, line 1)
- `TODO` next appears at byte offset 54 (body text, line 3)

### 3. Run Vale

```
vale test.md
```

**Expected:** Alert at line 3 (byte offset 54)
**Actual:** Alert at line 1, column 3 (first occurrence of "TODO")

## Root cause

In `AddAlert` (`internal/core/file.go`), for `scope: raw` blocks, the code always falls through to `FindLoc()` which performs a text search and returns the first occurrence. The byte offsets from `a.Span` are never consulted.

There is also a disambiguation attempt capped at 1000 characters:

```go
if len(a.Offset) == 0 && strings.Count(ctx, a.Match) > 1 && len(ctx) < 1000 {
    a.Offset = append(a.Offset, strings.Fields(ctx[0:a.Span[0]])...)
}
```

For `scope: raw`, `ctx` is the entire document, so any file over 1000 characters skips disambiguation entirely.

## Related

- #869: "Regex caret (^) not working properly for scope: raw when there are over 999 characters in a Markdown file" — directly caused by the 1000-char cap
- #272: "Exclusions don't seem to work properly in scope: raw"
- PR #1081: Proposed fix using a `HasByteOffsets` flag on the Alert struct

ryan-ronnander:
Fixes #1083

I found myself attempting to fix a raw "entire document" scope rule with a Tengo script. After working through the rule with Claude code, the alert's line and column values were not always correct. Claude discovered there was a bug in the upstream project. Here's the proposed fix.

## Summary

When a Tengo script rule uses `scope: raw`, Vale does not use the `begin` and `end` byte offsets returned by the script to calculate the alert's line and column. Instead, Vale extracts the matched text (`scope[begin:end]`) and performs a **text search** in the parsed document to determine the position. If the matched text appears multiple times in the document, Vale always reports the position of the **first** occurrence, regardless of which occurrence the script intended to flag.

## Vale version

v3.12.0 (still reproduces on v3.13.1)

### 1. Create a script rule

`styles/Example/FindTODO.yml`:

```yaml
extends: script
message: "Found a TODO."
scope: raw
level: warning
script: find-todo.tengo
```

`styles/config/scripts/find-todo.tengo`:

```tengo
text := import("text")
matches := []

// Hardcode a match at byte offset 54, which is the second
// occurrence of "TODO" in the test document (line 3).
matches = append(matches, {
    begin: 54,
    end:   58
})
```

### 2. Create a test document

`test.md` (the word "TODO" appears on lines 1 and 3):

```markdown
# TODO list for the project

This paragraph has a TODO that should be flagged by the script rule.
```

In this document:
- `TODO` first appears at byte offset 2 (in the heading, line 1)
- `TODO` next appears at byte offset 54 (in the body text, line 3)

### 3. Run Vale

```
vale test.md
```

### Expected behavior

The alert should be reported at **line 3** (where byte offset 54 falls), since the script returned `begin: 54, end: 58`.

### Actual behavior

The alert is reported at **line 1, column 3** — the position of the *first* occurrence of "TODO" in the document, not the occurrence at byte offset 54.

## Additional observations

- **Hardcoding `begin: 0, end: 1`** (matching just the first byte of the file) still reports the alert at `1:3` — confirming that byte offsets are completely ignored and Vale is searching for the extracted text.
- **Returning two matches** (both offset 2 and offset 54) correctly reports both `1:3` and `3:25` — because Vale finds the first and second occurrences in order.
- **Matching unique text** (text that only appears once in the document) always maps to the correct position, because the text search finds the right (only) occurrence.
- **Small files** appear to work correctly for common cases only because the matched text often happens to first appear at the intended location.

## Root cause

The bug flows through three files:

### 1. `internal/check/script.go` — Script execution (correct)

The `Run` method correctly extracts `begin`/`end` from the Tengo script output and sets `a.Span` and `a.Match` correctly.

### 2. `internal/core/file.go` — `AddAlert()` discards byte offsets

For `scope: raw`, `lintBlock` is called with `lookup=true`. Inside `AddAlert`, because `lookup=true`, the code **always** falls through to `FindLoc()`, ignoring the byte offsets the script provided.

There is also a disambiguation attempt capped at 1000 characters:

```go
if len(a.Offset) == 0 && strings.Count(ctx, a.Match) > 1 && len(ctx) < 1000 {
    a.Offset = append(a.Offset, strings.Fields(ctx[0:a.Span[0]])...)
}
```

For `scope: raw`, `ctx` is the **entire document**, so any file over 1000 characters skips this disambiguation entirely.

### 3. `internal/core/location.go` — `initialPosition()` searches for text

`FindLoc` delegates to `initialPosition()`, which builds a regex from `a.Match`, finds all occurrences, and **always returns the first one** — never consulting `a.Span` byte offsets.

## Fix

This PR adds a `HasByteOffsets` flag to the `Alert` struct and a `locFromByteOffset()` helper to compute line:column directly from byte offsets.

The approach:
1. **`internal/core/alert.go`** — A new `HasByteOffsets bool` field on `Alert` signals that `Span` contains byte offsets into the raw document (not column ranges).
2. **`internal/check/script.go`** — The script runner's `Run` method sets `HasByteOffsets: true` when building alerts from Tengo script matches, since scripts always return byte offsets.
3. **`internal/core/file.go`** — `AddAlert` checks `a.HasByteOffsets` to decide whether to use `locFromByteOffset()` (direct byte-offset-to-line:column conversion) or the existing `FindLoc` text-search path.

This is more precise than checking `blk.Context == blk.Text` to detect raw-scope blocks, which would also match non-script alerts (code comment blocks, plain text blocks) where `Span` contains column ranges rather than byte offsets.

The test expectation for `Scripts.CustomMsg` in `checks.feature` is updated from `4:19` to `1:2` — the old value was a side effect of the buggy text-search finding the matched text at the wrong position.

## Related issues

- #1083: Bug report for this issue
- #869: "Regex caret (^) not working properly for scope: raw when there are over 999 characters in a Markdown file" — directly caused by the 1000-char cap
- #272: "Exclusions don't seem to work properly in scope: raw"

### DIFF
--- a/internal/check/script.go
+++ b/internal/check/script.go
@@ -78,12 +78,13 @@ func (s Script) Run(blk nlp.Block, _ *core.File, _ *core.Config) ([]core.Alert, 
 		// don't use our custom regexp2 library, which means the offsets
 		// (`re2loc`) will be off.
 		a := core.Alert{
-			Check:    s.Name,
-			Severity: s.Level,
-			Span:     matchLoc,
-			Link:     s.Link,
-			Match:    matchText,
-			Action:   s.Action}
+			Check:          s.Name,
+			Severity:       s.Level,
+			Span:           matchLoc,
+			Link:           s.Link,
+			Match:          matchText,
+			Action:         s.Action,
+			HasByteOffsets: true}
 
 		if matchMsg, ok := match["message"].(string); ok {
 			a.Message, a.Description = formatMessages(matchMsg, s.Description, matchText)
--- a/internal/core/alert.go
+++ b/internal/core/alert.go
@@ -27,9 +27,10 @@ type Alert struct {
 	Message     string   // the output message
 	Severity    string   // 'suggestion', 'warning', or 'error'
 	Match       string   // the actual matched text
-	Line        int      // the source line
-	Limit       int      `json:"-"` // the max times to report
-	Hide        bool     `json:"-"` // should we hide this alert?
+	Line           int      // the source line
+	Limit          int      `json:"-"` // the max times to report
+	Hide           bool     `json:"-"` // should we hide this alert?
+	HasByteOffsets bool     `json:"-"` // Span holds byte offsets into the raw document
 }
 
 // FormatAlert ensures that all required fields have data.
--- a/internal/core/file.go
+++ b/internal/core/file.go
@@ -252,6 +252,32 @@ func (f *File) assignLoc(ctx string, blk nlp.Block, pad int, a Alert) (int, []in
 	return blk.Line + 1, a.Span
 }
 
+// locFromByteOffset computes a 1-based line number and a [col, col+len] span
+// from absolute byte offsets into the raw document text. This avoids the
+// text-search approach used by FindLoc/initialPosition, which can report the
+// wrong location when the matched text appears more than once.
+func locFromByteOffset(ctx string, begin, end, pad int) (int, []int) {
+	line := 1
+	lineStart := 0
+
+	for i := 0; i < begin && i < len(ctx); i++ {
+		if ctx[i] == '\n' {
+			line++
+			lineStart = i + 1
+		}
+	}
+
+	col := nlp.StrLen(ctx[lineStart:begin]) + 1 + pad
+	matchLen := nlp.StrLen(ctx[begin:end])
+
+	span := []int{col, col + matchLen - 1}
+	if span[1] <= 0 {
+		span[1] = 1
+	}
+
+	return line, span
+}
+
 // SetText updates the file's content, lines, and history.
 func (f *File) SetText(s string) {
 	f.Content = s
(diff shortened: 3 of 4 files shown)


jdkato:
Thanks!

PR #1096, opened 2026-04-21
Bugfix/path expansion

### LINKED ISSUE #1095: Global config that does not define StylesPath results in error with FindAsset
### Check for existing issues

- [x] Completed

### Environment

- Linux
- Go (local build)
- `vale version v3.14.1`

1. Using a user global config (`$HOME/.config/vale/.vale.ini`) and a project local config (`$PWD/.vale.ini`)
2. Define `StylesPath` in only the local configuration
3. Attempt to load an external style with a custom dictionary (`RedHat` for example)
4. The following error will be reported:
```
E201 Invalid value [.github/styles/RedHat/Spelling.yml:1:1]:

   1* ---
   2  extends: spelling
   3  level: warning

unable to resolve dicpath

Execution stopped with code 1.
```

`FindAsset` will be given the last `cfg.Path` in the list, which will not exist as it's a combination of the global config path with the local `StylePath`.

The way paths and configuration structures are handled right now is a little problematic. Paths should be expanded relative to the INI file before INI shadowing takes effect.

StaticRocket:
- fix(loadINI): use standard shadow loading

  Instead of appending configurations directly we should use the common shadow
  loading functions we've defined. This allows paths to be expanded properly
  ahead of time, and still use the shadow processing mechanisms other mechanisms
  expect.

  This also enforces the load order through the list of sources built and passed
  to processSources.

  Signed-off-by: Randolph Sapp <rs@ti.com>

- fix(ini): expand paths during load

  Expand INI values known to be paths during initial loading. This prevents
  having to guess where paths came from in later processing.

  Signed-off-by: Randolph Sapp <rs@ti.com>

Fixes: #1095

### DIFF
--- a/internal/core/ini.go
+++ b/internal/core/ini.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -12,6 +13,10 @@ import (
 	"github.com/errata-ai/vale/v3/internal/glob"
 	"github.com/errata-ai/vale/v3/internal/system"
 )
+
+var pathKeys = []string{
+	"StylesPath",
+}
 
 var coreError = "'%s' is a core option; it should be defined above any syntax-specific options (`[...]`)."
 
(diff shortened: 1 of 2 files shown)


jdkato:
Thanks!

PR #1110, opened 2026-06-28
fix(adoc): resolve empty-text cross-references corrupting walker context

### LINKED ISSUE #1109: Heading-scoped rules silently skip sections when a block anchor and an `xref:` macro coexist in AsciiDoc
### Check for existing issues

- [x] Completed

### Environment

- Vale 3.15.1
- Asciidoctor 2.0.26, resolved at `/opt/homebrew/bin/asciidoctor`
- Ruby 2.6.10p210
- macOS 26.5 on arm64
- Observed on 2026-06-19

Rules carrying `scope: heading` stop matching most section titles in an AsciiDoc file once that file combines explicit block anchors with `xref:` inline macros. The result is a silent false negative: a heading that violates the rule passes, because Vale never applies the rule to it.

Below is a self-contained reproduction, the expected and actual output, and two controls that each remove one half of the trigger and restore correct matching. The closing section relates this to the earlier issue #1101.

## Summary

A heading-scoped rule matches only a subset of section titles when both of these constructs are present in the same document:

- an explicit block anchor `[[id]]` on the line directly above a section title, and
- an `xref:` inline macro anywhere in the section bodies.

Removing either construct restores matching on every heading. Asciidoctor parses the same file with zero warnings, so the markup is valid and the defect sits in Vale's source-line mapping rather than in the document. A clean Vale run therefore does not prove that the headings conform.

## Minimal reproduction

Create three files in an empty directory.

`.vale.ini`:

```ini
StylesPath = .vale/styles
[*.adoc]
BasedOnStyles = English
```

`.vale/styles/English/TitleCase.yml`:

```yaml
extends: capitalization
message: "'%s' should use title-style capitalization."
level: error
scope: heading
match: $title
```

`repro.adoc`:

```asciidoc
= Document

[[intro]]
== first heading lowercase

This paragraph links to xref:#intro[].

[[middle]]
== second heading lowercase

This paragraph links to xref:#intro[].

[[final]]
== third heading lowercase

This paragraph links to xref:#intro[].
```

Run:

```console
$ vale repro.adoc
```

## Expected behavior

The rule carries `scope: heading` and each section title is lowercased, so Vale flags all three section titles. The expected output names lines 4, 9, and 14:

```
 4:4   error  'first heading lowercase' should use title-style capitalization.   English.TitleCase
 9:4   error  'second heading lowercase' should use title-style capitalization.  English.TitleCase
 14:4  error  'third heading lowercase' should use title-style capitalization.   English.TitleCase
```

## Actual behavior

Vale flags only the first section title and drops the other two:

```
 4:4  error  'first heading lowercase' should use title-style capitalization.  English.TitleCase

```

The section titles at lines 9 and 14 receive no `heading`-scoped match. Vale reports no error and exits as if those titles conform.

## The trigger is the pair of constructs

Two controls each remove one half of the trigger from `repro.adoc`. Each control restores matching on all three section titles.

Control one removes the `xref:` macros and keeps the anchors. Write the result to `ctrl_noxref.adoc` and lint it. All three section titles are flagged:

```console
$ sed 's/ to xref:#[a-z]*\[\]//' repro.adoc > ctrl_noxref.adoc
$ vale ctrl_noxref.adoc
 4:4   error  'first heading lowercase' should use title-style capitalization.   English.TitleCase
 9:4   error  'second heading lowercase' should use title-style capitalization.  English.TitleCase
 14:4  error  'third heading lowercase' should use title-style capitalization.   English.TitleCase

✖ 3 errors, 0 warnings and 0 suggestions in 1 file.
```

Control two removes the anchors and keeps the `xref:` macros. Write the result to `ctrl_noanchor.adoc` and lint it. All three section titles are flagged:

```console
$ grep -v '^\[\[' repro.adoc > ctrl_noanchor.adoc
$ vale ctrl_noanchor.adoc
 3:4   error  'first heading lowercase' should use title-style capitalization.   English.TitleCase
 7:4   error  'second heading lowercase' should use title-style capitalization.  English.TitleCase
 11:4  error  'third heading lowercase' should use title-style capitalization.   English.TitleCase

✖ 3 errors, 0 warnings and 0 suggestions in 1 file.
```

So neither construct alone causes the defect. The defect appears only when both are present.

## The document is valid AsciiDoc

Asciidoctor renders `repro.adoc` with no warnings and no errors:

```console
$ asciidoctor -o /dev/null repro.adoc
$ echo $?
0
```

The section titles, anchors, and cross-references are well-formed. The mismatch arises after the render, when Vale maps rendered spans back to source line numbers.

## Impact

The defect is a silent false negative on every heading-scoped rule, not Title Case alone. Any rule with `scope: heading` is suppressed on the dropped titles. A team that relies on Vale to enforce heading conventions receives a passing run while violations remain in the dropped titles.

The two triggering constructs are common in long technical documents. Explicit block anchors and `xref:` cross-references are the standard AsciiDoc mechanism for stable internal links, and some house style guides require them. Such documents are the ones most likely to depend on automated heading checks, and they are the ones where the checks fail.

## Related issue

This defect resembles issue #1101, in which heading-scoped rules did not match AsciiDoc headings at all. That issue is marked fixed as of release 3.15.0. The reproduction in this report runs on 3.15.1, which is past that fix. Title Case matches a heading in a trivial file, which confirms the 3.15.0 fix is active, yet the pairing of block anchors with `xref:` macros still desynchronizes the source-line mapping and drops the heading scope. This report therefore documents a distinct manifestation that survives the 3.15.0 fix rather than a regression of it.

SudhanshuMatrix:
## Description
This PR fixes a silent false-negative bug where rules carrying `scope: heading` stop matching section titles in an AsciiDoc file if that file combines explicit block anchors with `xref:` inline macros (having empty link text).

Fixes #1109

### Root Cause
1. **Dynamic Target Resolution**: During HTML generation, Asciidoctor resolves empty-text cross-references (like `xref:id[]` or `<<id>>`) using the target heading's title.
2. **Missing Source Text**: Since the resolved text is not written in the raw source paragraph containing the reference, Vale's walker fails to locate the exact phrase inside the paragraph block.
3. **Word-Masking Fallback**: As a fallback, Vale starts masking the individual words (`"first"`, `"heading"`, `"lowercase"`) anywhere it can find them in the remaining document context.
4. **Context Corruption**: This matches and masks the words inside subsequent headings (e.g., `"second heading lowercase"` becomes `"second @@@@@@@ @@@@@@@@@"`).
5. **Silent Drop**: When Vale later processes these headings, they fail to match the corrupted context, leading to location lookup failures and the alert being silently discarded.

## Proposed Changes
We intercept the AsciiDoc content inside `lintADoc` in `internal/lint/adoc.go` before it is passed to Asciidoctor, and rewrite empty-text references:
- `xref:target[]` -> `xref:target[target]`
- `<<target>>` -> `<<target,target>>`

This ensures that the text rendered in the HTML tag is the literal target identifier (e.g. `#intro` or `intro`), which is present in the raw source code of the paragraph. Vale's walker successfully matches and masks it within the paragraph block, leaving subsequent headings untouched.

## How to Test and Verify

1. Run Vale with a heading-scoped rule against this sample AsciiDoc file (`repro.adoc`):
   ```asciidoc
   = Document

   [[intro]]
   == first heading lowercase

   This paragraph links to xref:#intro[].

   [[middle]]
   == second heading lowercase

   This paragraph links to xref:#intro[].

   [[final]]
   == third heading lowercase

   This paragraph links to xref:#intro[].

### DIFF
--- a/internal/lint/adoc.go
+++ b/internal/lint/adoc.go
@@ -25,6 +25,8 @@ var adocSanitizer = strings.NewReplacer(
 // Convert listing blocks of the form `[source,.+]` to `[source]`
 var reSource = regexp.MustCompile(`\[source,.+\]`)
 var reComment = regexp.MustCompile(`// .+`)
+var reXref = regexp.MustCompile(`xref:([^\[]+)\[\]`)
+var reShortXref = regexp.MustCompile(`<<([^>,]+)>>`)
 
 var adocArgs = []string{
 	"-s",
@@ -63,6 +65,9 @@ func (l *Linter) lintADoc(f *core.File) error {
 		return err
 	}
 	s = adocSanitizer.Replace(s)
+
+	s = reXref.ReplaceAllString(s, "xref:$1[$1]")
+	s = reShortXref.ReplaceAllString(s, "<<$1,$1>>")
 
 	html, err = callAdoc(s, exe, l.Manager.Config.Asciidoctor)
 	if err != nil {


jdkato:
Thanks for the PR and the root-cause writeup. The diagnosis is right: empty AsciiDoc xrefs render to the target heading text, that rendered text is not present in the source paragraph, and Vale’s word-by-word fallback can then mask words in later headings. That is what causes the heading-scoped false negatives.

I don’t think we should fix this by rewriting the AsciiDoc before render. `xref:#intro[]` and `xref:#intro[#intro]` are not equivalent from Vale’s point of view: the former renders as the target heading title, while the latter renders as `#intro`. That means paragraph/link-adjacent rules would lint text that Asciidoctor would not actually render. This fixes the heading bug by changing the document Vale analyzes, which is the wrong layer.

I tested a smaller fix in the walker instead: `internal/lint/walk.go` (https://github.com/vale-cli/vale/commit/778a44f69507e0400c8933640600a613f145cf02)  should preserve Asciidoctor’s rendered output and avoid the global word-by-word masking fallback for `.adoc` when rendered text is not present verbatim in the source. That prevents synthetic xref labels from corrupting later headings without changing the prose Vale sees.

Issue #1052, opened 2025-11-04
Two words split by code block in MDX


verhovsky:
### Check for existing issues

- [x] Completed

### Environment

- macOS
- Homebrew
- `vale version 3.13.0`

### Describe the bug / provide steps to reproduce it

This file:

```mdx
in `$VARIABLE` for.

information
```

when parsed as MDX with this .vale.ini

```ini
SkippedScopes = code

[formats]
mdx = md

[*.mdx]
BasedOnStyles = Vale
```

Gives this error:

```
 test.mdx
 3:1  error  Did you really mean 'infor'?  Vale.Spelling 

✖ 1 error, 0 warnings and 0 suggestions in 1 file.
```

There should be no error because "in" "for" and "information" are all normal words. For some reason having "in" and "for" separated with a `code` block and then followed by a word that uses both words causes an error on that word (notice the error is on line 3, not 1).

colinloretz:
I am seeing this as well across multiple words where it is detecting tokens inside full words.


Specifically for the `infor` one, the line is:

```markdown
The interaction payload will also include a new `attachment_size_limit` key that specifies the maximum allowed attachment size. This limit may be higher than the default attachment size limit, depending on the guild's boost status or the invoking user's Nitro status.

For more information, check out [our documentation on file uploads](/developers/docs/reference#uploading-files).
```

When tested in isolation, the same string passes fine:

```sh
echo "For more information, check out [our documentation on file uploads](/developers/docs/reference#uploading-files)." | vale --ext=.mdx
```

Other examples of false positives:

```sh
 developers/docs/developer-tools/game-sdk.mdx
 157:238  error  Did you really mean 'inin'?    Vale.Spelling 
 274:27   error  Did you really mean 'callin'?  Vale.Spelling 


 developers/docs/events/gateway.mdx
 304:184  error  Did you really mean 'passin'?  Vale.Spelling 
 609:39   error  Did you really mean 'asin'?    Vale.Spelling 


 developers/docs/interactions/overview.mdx
 162:239  error  Did you really mean 'ato'?  Vale.Spelling 


 developers/docs/resources/channel.mdx
 553:48  error  Did you really mean 'bein'?  Vale.Spelling 

 developers/docs/tutorials/developing-a-user-installable-app.mdx
 358:220  error  Did you really mean 'inin'?  Vale.Spelling 

 developers/docs/change-log.mdx
 745:12  error  Did you really mean 'infor'?  Vale.Spelling 
```

Issue #1001, opened 2025-05-01
Vale off/on comments do not work in Markdown on a line containing a bold or italic word


dagardner-nv:
### Check for existing issues

- [x] Completed

### Environment

* OS: Linux, Debian 12.10
* Install method Conda
* 3.11.2 and 3.9.5 (Originally observed with version 3.9.5, reproduced with 3.11.2)

### Describe the bug / provide steps to reproduce it

A word (`TODO`) listed in the `reject.txt` file is being flagged as an error despite being surrounded by `vale off` and `vale on` comments in a Markdown file. 

The issue appears to occur if the word appears on the same line as a bold or italicized word.

ex:
```
*Note*: <!-- vale off -->TODO<!-- vale on -->
```

This is also flagged:
```
**Note**: <!-- vale off -->TODO<!-- vale on -->
```

as are these:
```
_Note_: <!-- vale off -->TODO<!-- vale on -->
__Note__: <!-- vale off -->TODO<!-- vale on -->
```

However this works as expected:
```
Note: This works as expected <!-- vale off -->TODO<!-- vale on -->
```

Full repro is available at: https://github.com/dagardner-nv/vale-off-test

Example output:
```
$ vale ./README.md 

 README.md
 5:26   error  Avoid using 'TODO'.  Vale.Avoid 
 7:26   error  Avoid using 'TODO'.  Vale.Avoid 
 9:28   error  Avoid using 'TODO'.  Vale.Avoid 
 11:28  error  Avoid using 'TODO'.  Vale.Avoid 

✖ 4 errors, 0 warnings and 0 suggestions in 1 file.
```

Possibly related to issues #340 & #878 however those were different enough that I decided to open an issue.

PR #1140, opened 2026-08-14
fix(lint): treat an HTML comment as a text boundary outside inline elements

### LINKED ISSUE #882: Comments in markdown trigger erroneous spelling error
### Check for existing issues

- [X] Completed

### Environment

Reproducible on:

- MacOS, installed via homebrew, vale version 3.6.1
- MacOS, GitHub binary, vale version 3.7.0

With the default config:

```
MinAlertLevel = suggestion

[*]
BasedOnStyles = Vale
```

This test document:

```
on <!-- TODO --> a 
sonata
```

Triggers a spelling error:

```
$ ./vale test.md 

 test.md
 2:2  error  Did you really mean 'ona'?  Vale.Spelling 

```

My expectation was no errors would be reported.

Remove the comment in the test document and no errors are reported.

The pattern seems to be that the non-comment characters on the first line, in this case `on` and `a`, will be flagged as a spelling error on their first usage after the comment. If you break the HTML comment, like this `< !-- TODO -->`, the unexpected behavior does not occur.

Another failing example is:

```
on <span/> a 
sonata
```

By setting `BlockIgnores = (?s) (<!--.*?-->)|(<[^>]+\/>)` will workaround this issue. This does not, however, explain the unusual behavior where a portion of a valid word is flagged as a spelling error based on the preceding line.

VXNCXNX:
An HTML comment between two words causes a false spelling alert:

```
$ cat test.md
on <!-- TODO --> a
sonata

$ vale test.md
test.md:2:2  error  Did you really mean 'ona'?  Vale.Spelling
```

Removing the comment yields no alerts.

`lintHTMLTokens` sets `closedInline` in the `EndTagToken` branch and clears it in `StartTagToken`, but a `CommentToken` leaves it untouched. Its consumer is `clean()`, whose whitespace decision is `(closedInline && spaced) || (!closedInline && inline && !starter)`. After a comment both flags are stale-false, so neither arm fires, the following text loses its leading space, and `on` + `a` fuse into `ona`.

Setting `closedInline` on a comment only fixes the outer case: a comment right after an inline *opening* tag is not a closed boundary, and asserting one there disables the `#1052` padding. That loses real alerts and invents words — `A sona<em><!--c-->ta</em>` fuses to `sonata`, so the genuine `sona` alert disappears.

So the flag is set only when no inline element is open. Inside one, the opening tag remains the boundary and its padding is preserved.

Fixes #882

## Verification

Built before/after binaries and compared:

| input | before | after |
|---|---|---|
| `on <!-- TODO --> a` / `sonata` | `2:2 'ona'` | no alert |
| `A sona<em><!--c-->ta</em> here.` | `1:3 'sona'` | `1:3 'sona'` |
| `A sona<b><!-- c --></b>ta here.` | unchanged | unchanged |
| `in <code><!--c-->x</code> for` | unchanged | unchanged |
| `word<!--c-->s` | stays `words` | stays `words` |
| `A <!-- c --> sona` | no alert (fused to `Asona`) | `1:14 'sona'` |

The last row is a second bug the same change fixes: text after a comment at the start of a fragment was being fused to the preceding word.

`go build ./...` and `go test ./...` pass. The regression case is in `testdata/e2e/comments.yaml` alongside the existing `#1001` comment case, and it fails without the fix.

Note `lintHTMLTokens` is shared across html/md/adoc/rst/org/dita/xml, so this affects all of them; I checked the `.html` path behaves the same.

### DIFF
--- a/internal/lint/ast.go
+++ b/internal/lint/ast.go
@@ -163,6 +163,13 @@ func (l *Linter) lintHTMLTokens(f *core.File, raw []byte, offset int) error { //
 				}
 			}
 		} else if tokt == html.CommentToken {
+			// A comment, like a closed inline element, is a source-faithful
+			// boundary: trust the next text's own whitespace (#882). Inside a
+			// still-open inline element the opening tag remains the boundary,
+			// so leave the state alone and keep its padding (#1052).
+			if !inline {
+				closedInline = true
+			}
 			f.UpdateComments(txt)
 			walker.update(txt, tokt)
 		} else if tokt == html.TextToken {
--- a/testdata/e2e/comments.yaml
+++ b/testdata/e2e/comments.yaml
@@ -100,3 +100,10 @@ cases:
     exit: 0
     want: |
       issue1001.md:5:28:demo.Ending-Preposition:Don't end a sentence with 'of.'
+
+  - name: inline-comment-spacing
+    about: "#882 -- text after a comment keeps its own leading space instead of
+      fusing with the text before it"
+    args: issue882.md
+    exit: 0
+    want: ""
--- a/testdata/e2e/comments.yaml
+++ b/testdata/e2e/comments.yaml
@@ -103,7 +103,9 @@ cases:
 
   - name: inline-comment-spacing
     about: "#882 -- text after a comment keeps its own leading space instead of
-      fusing with the text before it"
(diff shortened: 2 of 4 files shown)


jdkato:
Thanks!

Issue #1111, opened 2026-06-28
Heading checks false-positive when inline code is followed by an `++...++` passthrough (AsciiDoc)


ciechanowiec:
### Check for existing issues

- [x] Completed

### Environment

- Vale 3.15.1
- macOS (Darwin 25.5.0)
- Format: AsciiDoc (`.adoc`)

## Summary

In an AsciiDoc heading, when an inline code span is immediately followed by an inline passthrough (`` `code`++...++ ``), the text Vale extracts for heading-scoped checks is corrupted: the code span is replaced by a run of `*`, and the passthrough content is split off as a separate, space-prefixed token. A check scoped to headings then runs against this mangled text and reports a false positive.

I hit this with a `capitalization` / `match: $title` rule, but the corruption is in the extracted text itself, so any heading-scoped check is affected.

## Steps to reproduce

Three files:

`.vale.ini`

```ini
StylesPath = styles

[*.adoc]
BasedOnStyles = Test
```

`styles/Test/Heading.yml`

```yaml
extends: capitalization
message: "'%s' should use title-style capitalization."
level: error
scope: heading
match: $title
```

`test.adoc`

```asciidoc
= T

==== Mixing `ResourceProvider`++s++

==== One Tree of `Resource`++s++

==== Using `Foo`++s++ Here
```

Run:

```sh
vale test.adoc
```

## Actual behavior

```
 test.adoc
 3:1  error  'Mixing **************** s' should use title-style capitalization.  Test.Heading
 7:1  error  'Using *** s Here' should use title-style capitalization.           Test.Heading
```

Two things are visible in the messages themselves:

- The inline code span is masked as `*` (16 for `ResourceProvider`, 3 for `Foo`), which is expected - Vale skips code.
- The passthrough `++s++` is detached as a separate ` s` token with a leading space, instead of being attached to the code span as `ResourceProviders`. That stray lowercase `s` is what the title-case check trips over.

## Expected behavior

No error. `` `ResourceProvider`++s++ `` renders as `ResourceProviders`, and "Mixing `ResourceProviders`" is valid title case. The extracted text should keep the `s` attached to the (masked) code span rather than splitting it into its own token.

## What makes me think it's a tokenization bug, not the rule

[...]

PR #1142, opened 2026-08-15
fix(lint): apply suppression comments to relocated footnote definitions

### LINKED ISSUE #1078: Vale suppression comments don't work for footnotes.
### Check for existing issues

- [x] Completed

### Environment

MacOS installed via Homebrew.
Also GitHub runners with `ubuntu-latest` installed via [`vale-action`](https://github.com/errata-ai/vale-action)

```
❯ vale --version
vale version 3.13.1
```

This seems to have been reported in #927 which was closed (?) 

`mwe.md`
```markdown
Here is some text that has a footnote[^footnote-link].

<!-- vale proselint.Annotations = NO -->

[^footnote-link]: I want to keep Proselint-dot-Annotations enabled globally to flag up FIXMEs and TODOs. But! Annoyingly, there is a legitimate group called the [TODO Group](https://todogroup.org/) and I want to talk about them in a footnote.

<!-- vale proselint.Annotations = YES -->

Some more text, let's pretend that this is a longer document with other sections and other footnotes
```

`vale.ini`
```ini
StylesPath = .styles

# https://github.com/errata-ai/packages
Packages = proselint

[*.md]
BasedOnStyles = proselint
```

The comment suppressions turning off and on seem to break in chunks of footnotes...

```
❯ vale sync && vale mwe.md
Syncing proselint [1/1] █████████████████████████████████████████████ 100% | 0s
 SUCCESS  Synced 1 package(s) to '/Users/scnlf/projects/open-source-seed/mwe-vale/.styles'.

 mwe.md
 6:162  error  'TODO' left in text.  proselint.Annotations

```
Which should be suppressed by the comments.

Have also tried `<!-- vale off -->`/`<!-- vale on -->` with the same results.

VXNCXNX:
Fixes #1078.

## What's broken

Suppression comments do not suppress alerts inside footnote definitions.

```
$ cat mwe.md
Here is some text that has a footnote[^x].

<!-- vale Vale.Spelling = NO -->

[^x]: I want to talk about thingamajigg in a footnote.

<!-- vale Vale.Spelling = YES -->

Some more text here about nothing.

$ vale --output=line mwe.md
mwe.md:5:40:Vale.Spelling:Did you really mean 'thingamajigg'?
```

The same document with the footnote replaced by a plain paragraph reports nothing.

## The cause

Not what the issue reporter guessed. goldmark's Footnote extension relocates every definition into a trailing `div.footnotes`, which I confirmed by dumping the rendered HTML:

```
<!-- vale Vale.Spelling = NO -->
<!-- vale Vale.Spelling = YES -->
<p>Some more text here about nothing.</p>
<div class="footnotes" ...>
<li id="fn:1"><p>I want to talk about thingamajigg ...</p></li>
```

The toggle state is a linear walk over the token stream, so by the time the relocated text is reached, the `= YES` has already re-enabled the check. The telling detail is the converse: an unterminated `= NO` does suppress the footnote, which is only possible if position in the stream, not the footnote itself, is the variable.

## The fix

Inside that div, rebuild the state from the definition's own source line by replaying the comment controls that precede it, then restore the ambient state on the closing tag.

One deviation worth calling out: I key on the definition's source position, not the reference's. The issue's own example has the reference on line 1, outside the suppression block, with only the definition inside it. Keying on the reference leaves that example still reporting the alert. I built it that way first and confirmed it.

## MyST and qmd

The `div.footnotes` container is a goldmark promise, not a general one, so the branch is gated on `.md`, `.mdx`, `.myst` and `.qmd`. Those are exactly the four whose converters register `extension.Footnote`, so rST, AsciiDoc and HTML never enter it. MyST and qmd are included because they relocate footnotes too and had the same bug; their existing tests pass unchanged.

## Verification

`go build ./...` clean and `go test ./internal/...` passes in full.

The MWE above now reports nothing and exits 0. Deleting the suppression comments brings the alert back.

The interesting case is this one, which broke my first attempt:

```
An intro mentioning thingamajigg early on, plus a footnote[^x].

<!-- vale Vale.Spelling = NO -->

[^x]: I want to talk about thingamajigg in a footnote.
```

Resolving the definition by searching the whole document for its longest word finds line 1, so the replay stopped before the `= NO` and the definition was not suppressed. It now walks the source's definition starts and picks the one whose body contains the word. Only the line-1 alert is reported, which is correct.

Also checked: two footnotes where one is suppressed and one is not, and a definition whose longest word recurs in a later paragraph. Both behave.

One honest weakness. `f.Content` is `prepMarkdown`-processed, which stars out the label, so `[^x]:` becomes `[**]:` and the regex has to match definitions by shape rather than by `\[\^`. That means a link reference definition could in principle be selected first if it contained the same longest word. I could not construct a case where this produced a wrong suppression, since link definition labels are starred out and their URLs are not linted, but the ambiguity is there and I would rather you knew about it than found it.

This touches the same walker branch as #1001, #1052 and #1111. The change is purely additive, three touch points in the loop and a self-contained type below `lintHTMLTokens`, so it should not conflict badly.

### DIFF
--- a/internal/lint/ast.go
+++ b/internal/lint/ast.go
@@ -2,6 +2,7 @@ package lint
 
 import (
 	"bytes"
+	"regexp"
 	"strings"
 	"unicode/utf8"
 
@@ -45,6 +46,13 @@ var inlineToScope = map[string]string{
 	"tt":     "code",
 }
 
+// goldmarkFootnotes names the formats goldmark converts. Its Footnote
+// extension moves every definition into a trailing `div.footnotes`, so the
+// comment state the walker reaches there is whatever the end of the document
+// left behind rather than what governed the reference. Elsewhere -- rST,
+// AsciiDoc, HTML -- that class carries no such promise. See #1078.
+var goldmarkFootnotes = []string{".md", ".mdx", ".myst", ".qmd"}
+
 var tagToScope = map[string]string{
 	"th":         "text.table.header",
 	"td":         "text.table.cell",
@@ -87,9 +95,16 @@ func (l *Linter) lintHTMLTokens(f *core.File, raw []byte, offset int) error { //
 	}
 	var open []inlineCapture
 
+	var notes footnotes
+	relocated := core.StringInSlice(f.NormedExt, goldmarkFootnotes)
+
 	walker := newWalker(f, raw, offset)
 	for {
 		tokt, tok, txt := walker.walk()
+
+		if relocated {
+			notes.enter(f, tokt, tok, txt)
+		}
 
 		walker.addCls(txt, tokt == html.StartTagToken)
 		closed := walker.canClose()
@@ -250,6 +265,9 @@ func (l *Linter) lintHTMLTokens(f *core.File, raw []byte, offset int) error { //
 		if tokt == html.EndTagToken && !core.StringInSlice(txt, inlineTags) {
 			content := buf.String()
 			if strings.TrimSpace(content) != "" {
+				if notes.depth > 0 {
+					notes.apply(f, content)
+				}
 				err := l.lintScope(f, walker, content)
 				if err != nil {
 					return err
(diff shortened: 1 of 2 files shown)


jdkato:
Thanks — the diagnosis is right, but I don't want to take this mechanism. It adds a second text-to-source heuristic beside the one the walker already uses for alert positions, mutates `f.Comments` mid-walk, and hardcodes goldmark's `div.footnotes` into the format-agnostic walker.

Issue #878, opened 2024-07-27
Vale on/off comments don't work with displayed equation


datawookie:
### Check for existing issues

- [X] Completed

### Environment

Ubuntu 23.04
Snap

### Describe the bug / provide steps to reproduce it

Hi!

Firstly, thanks for the work on Vale. I'm using it with my blog and it has helped me to identify countless errors in my prose. :pray: 

I have come upon something puzzling. In a Markdown file I have the following:

```
<!-- vale off -->
Consider some time series data consisting of \\(n\\) points, \\(y_t = x_t + \epsilon_t\\), where \\(x_t\\) is a smooth signal, \\(\epsilon_t\\) is noise and \\(y_t\\) is the combined noisy sign. If we have observations of \\(y_t\\), how can get back to an estimate of \\(x_t\\)?
<!-- vale on -->

$$\frac{1}{2} \sum\_{t=1}^n (y\_t - x\_t)^2 + \lambda \sum\_{t=2}^{n-1} (x\_{t-1} - 2 x\_t + x\_{t+1})^2.$$
```

In this case the on/off comments don't work and the spelling of `x_t` and `y_t` is highlighted.

If, however, I remove the displayed equation (as below) then the comments work fine.

```
<!-- vale off -->
Consider some time series data consisting of \\(n\\) points, \\(y_t = x_t + \epsilon_t\\), where \\(x_t\\) is a smooth signal, \\(\epsilon_t\\) is noise and \\(y_t\\) is the combined noisy sign. If we have observations of \\(y_t\\), how can get back to an estimate of \\(x_t\\)?
<!-- vale on -->
```

Thanks, Andrew.

jdkato:
`$$…$$` display math is now parsed as math and skipped (not spell-checked), so equations no longer produce prose findings.

```markdown
<!-- vale off -->
Consider \\(x_t\\), \\(y_t\\) …
<!-- vale on -->

$$\frac{1}{2} \sum\_{t=1}^n (y\_t - x\_t)^2 …$$
```
-> no spelling errors.

Two notes:
- This covers `$$…$$` blocks. Inline single-`$` math (`$x$`) is intentionally **not** treated as math, so prose with currency like `$5 and $10` keeps getting linted normally.
- A `$$` block must be closed (an unclosed `$$` consumes the rest of the document, same as an unclosed ``` ``` ``` fence).

PR #1149, opened 2026-08-18
fix(lint): close display math on more than a bare $$ line

### LINKED ISSUE #1148: Markdown: display math only closes on a bare `$$` line, so many `$$…$$` forms silently swallow the rest of the file
### Check for existing issues

- [x] Completed

### Environment

- **OS**: macOS 26.5.2 (arm64)
- **Install method**: Homebrew
- **Vale version**: `vale version 3.17.1`

### Summary

Since v3.15.2, `$$…$$` display math is skipped when linting Markdown (#878). The block is only ever *closed*, though, by a line that contains nothing but `$$`. A closing `$$` anywhere else on a line is not recognised, the block is treated as unclosed, and — as noted on #878, an unclosed block consumes the rest of the document — everything after it goes unlinted.

The shapes that break are ordinary Pandoc display math, so this is not an edge case:

- `$$x=1$$` — opening and closing on one line
- `$$\begin{aligned} … \end{aligned}$$` — closing delimiter ending a content line
- `$$ … $$ {#eq-foo}` — a Pandoc/Quarto cross-reference label after the closing delimiter

The opening delimiter is fine either way; `$$ x=1` on the opening line works as long as the closer sits alone. So the rule in practice is *the closing `$$` must be the only thing on its line*, which is stricter than Pandoc.

This is easy to miss because the affected file reports **zero alerts** rather than an error. In my own corpus (a Quarto site, `qmd = md` under `[formats]`), 275 files contain `$$` and 85 of them were reporting zero alerts and reading as clean. Masking the math surfaced 1,530 genuine warnings across 179 files.

### Steps to reproduce

`.vale.ini` — built-in style only, no `StylesPath` contents needed, no `[formats]` mapping:

```ini
StylesPath = styles
MinAlertLevel = suggestion

[*.md]
BasedOnStyles = Vale
```

Four documents, each with the same misspelling before and after the math:

```bash
printf 'A speling error before the math.\n\n$$\nx=1\n$$\n\nAnother speling error after it.\n'                       > own-line.md
printf 'A speling error before the math.\n\n$$x=1$$\n\nAnother speling error after it.\n'                            > single-line.md
printf 'A speling error before the math.\n\n$$\nx=1\n$$ {#eq-foo}\n\nAnother speling error after it.\n'              > labelled.md
printf 'A speling error before the math.\n\n$$\\begin{aligned}\nx &= 1\n\\end{aligned}$$\n\nAnother speling error after it.\n' > env.md

vale --output=line own-line.md single-line.md labelled.md env.md
```

### Actual

Only `own-line.md` reports the second misspelling. The other three stop at the math:

```
own-line.md:1:3:Vale.Spelling:Did you really mean 'speling'?
own-line.md:7:9:Vale.Spelling:Did you really mean 'speling'?
single-line.md:1:3:Vale.Spelling:Did you really mean 'speling'?
labelled.md:1:3:Vale.Spelling:Did you really mean 'speling'?
env.md:1:3:Vale.Spelling:Did you really mean 'speling'?
```

### Expected

All four report both misspellings — two alerts per file, eight in total. All four are display math to Pandoc, and in each the prose after the equation is prose.

### Cross-check against Pandoc

Each of the four parses to a single `DisplayMath` node under `pandoc 3.10.1`:

```bash
printf '$$x=1$$\n\nAfter.\n' | pandoc -f markdown -t json | grep -o DisplayMath
```

So the divergence is in what counts as a closing delimiter, not in what counts as math. The Markdown docs say display math is ignored and that inline math follows Pandoc's delimiter rules; display math currently follows a stricter rule than either.

### Suggested behaviour

Recognise a closing `$$` that is not alone on its line — at minimum the single-line and end-of-line forms — and, since an unclosed block is documented as consuming the remainder of the document, consider whether that case is worth surfacing rather than leaving silent. A blank line is a reasonable bound: Pandoc will not carry display math across one either (jgm/pandoc#5669).

### A test case, in the `internal/e2e` format

```yaml
- name: math/closing-delimiter
  about: >-
    A closing `$$` need not be alone on its line. Each of these is display math
    to Pandoc, so the prose after it is still prose. See #878.
  files:
    .vale.ini: |
      StylesPath = styles
      MinAlertLevel = suggestion

      [*.md]
      BasedOnStyles = Vale
    single-line.md: |
      A speling error before the math.

      $$x=1$$

      Another speling error after it.
    labelled.md: |
      A speling error before the math.

      $$
      x=1
      $$ {#eq-foo}

      Another speling error after it.
  args: single-line.md labelled.md
  exit: 1
  want: |
    labelled.md:1:3:Vale.Spelling:Did you really mean 'speling'?
    labelled.md:7:9:Vale.Spelling:Did you really mean 'speling'?
    single-line.md:1:3:Vale.Spelling:Did you really mean 'speling'?
    single-line.md:5:9:Vale.Spelling:Did you really mean 'speling'?
```

### Workaround, for anyone who finds this

Masking the math in `.vale.ini` works, but it needs both keys rather than one:

```ini
TokenIgnores = (?m)^ *(\$\$.*\$\$)
BlockIgnores = (?m)^ *(\$\$(?:.|[[:space:]].)*?[[:space:]]?[[:blank:]]*\$\$)
```

Both patterns run over raw text, so neither can see a code fence. A *block* substitution landing inside a fence terminates it and hands the remainder of the file over as code — which silently loses alerts on any page that quotes display-math syntax rather than using it. `TokenIgnores` substitutes inline, a fence survives it, and it is applied first, so giving it every single-line `$$…$$` leaves `BlockIgnores` only the multi-line spans. Two other things that cost me time: the INI reader expands `\n` inside a value into a real newline, which corrupts the config, so the patterns are written in POSIX classes; and a `(?s)`-style body was pathologically slow — 30s on one 400-line file, against 3s for a 1,725-file corpus with the patterns above.

----
AI Disclosure: diagnosis and repro done for me by Claude, but I'm running Vale right now with my human eyeballs and it seems correct.

ChrisJr404:
Recognize every place Pandoc closes a `$$…$$` display-math block, so an equation no longer leaves the block open and hides the rest of the file from linting.

Since v3.15.2 (#878) display math is parsed by goldmark-mathjax's block parser, which only closes the block on a line that contains nothing but `$$`. That misses the ordinary Pandoc shapes reported in #1148:

- `$$x=1$$` — opener and closer on one line
- `$$\begin{aligned} … \end{aligned}$$` — the closing `$$` ends a content line
- `$$ … $$ {#eq-foo}` — a Pandoc/Quarto cross-reference label after the close

In each case the block stays open, and an open block consumes everything to the end of the document, so all the prose past the equation goes unchecked.

Rather than patch around the third-party parser, this replaces it with an in-tree block parser — the same move already made for inline math, and for the reason the file's comment already gives. The new parser produces the same `mathjax.MathBlock` node the renderer knows how to skip, keeps a bare `$$` line as markup exactly as before, and additionally closes on an opener that also closes, on a content line ending in `$$`, and after an optional trailing `{…}` label.

### Testing

`internal/lint/math_test.go` gains `TestMathBlock`, which runs each shape above (plus the existing bare-`$$` block, to pin that it still behaves) through the Quarto renderer and asserts both that the equation renders as skipped `pre` and that the paragraph after it survives as prose. `go test ./internal/lint/` and `golangci-lint run ./internal/lint/` both pass.

Closes #1148.

### DIFF
--- a/internal/lint/math.go
+++ b/internal/lint/math.go
@@ -1,6 +1,8 @@
 package lint
 
 import (
+	"bytes"
+
 	"github.com/yuin/goldmark"
 	"github.com/yuin/goldmark/ast"
 	"github.com/yuin/goldmark/parser"
@@ -15,14 +17,17 @@ import (
 // them as elements vale skips -- so equations aren't spell-checked as prose
 // (#878, #839).
 //
-// The inline parser is ours rather than goldmark-mathjax's: that one takes any
-// two `$` on a line as delimiters, so `$5 and $10` reads as math and the prose
-// between them is silently dropped from linting. See mathInlineParser.
+// Neither parser is goldmark-mathjax's. Its inline parser takes any two `$` on
+// a line as delimiters, so `$5 and $10` reads as math and the prose between
+// them is silently dropped (see mathInlineParser). Its block parser only ever
+// closes on a line that is nothing but `$$`, so `$$x=1$$`, a content line that
+// ends in `$$`, and a `$$ … $$ {#eq-foo}` label all leave the block open -- and
+// an unclosed block consumes the rest of the file (see mathBlockParser, #1148).
 type mathExtension struct{}
 
 func (mathExtension) Extend(m goldmark.Markdown) {
 	m.Parser().AddOptions(parser.WithBlockParsers(
-		util.Prioritized(mathjax.NewMathJaxBlockParser(), 701),
+		util.Prioritized(mathBlockParser{}, 701),
 	))
 	m.Parser().AddOptions(parser.WithInlineParsers(
 		// '$' has no built-in owner.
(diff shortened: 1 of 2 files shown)


jdkato:
Thanks — this is the right direction, but there are a few issues.

1. It panics on back-to-back blocks (`$$\nx\n$$\n$$\ny\n$$`) and on an escaped `\$$` before a bare `$$`. The context-key state is the cause: goldmark opens the next block *before* calling `Close` on the previous one, so the old block's `Close` nils the key the new block just set. 
2. An unclosed `$$` still swallows the rest of the file — the blank-line bound from #1148
   isn't implemented.

I've got a branch covering both plus e2e tests, so I'll close this in favor of that (2f65b836c2ad3dc92d6df0c9f65174af382d9ee0) — appreciate the report regardless.

PR #1151, opened 2026-08-21
feat: add Elixir comment and doc-attribute extraction

C-Sinclair:
Adds Elixir (`.ex`, `.exs`) to the tree-sitter comment extractors.

### Why the doc attributes, not just the comments

Elixir has no documentation comment syntax. Its published API documentation lives in module attributes holding a string or a heredoc:

```elixir
defmodule Session do
  @moduledoc """
  A scheduled period of care delivery.
  """

  @doc "Books a session for a client."
  def book(client), do: ...
end
```

That is what `mix docs` renders and what a reader of a module reads first, so extracting only `#` comments would see the asides and none of the documentation.

Both are extracted. Comments keep the usual scopes; the attributes get a `doc` meta scope, so `text.comment.doc.block` can be held to a different standard than an implementation note, or excluded on its own via `IgnoredScopes`. `@moduledoc`, `@doc`, `@typedoc` and `@shortdoc` are covered; `@doc false` and `@doc since: "1.0.0"` carry no prose and the queries require a string or sigil argument, so neither matches.

### Predicate-only captures

The attribute name is what decides whether a node is documentation, and the prose is a *different* node — so the query has to test one node and capture another. `#match?` can only test a captured node, so this needs a capture that is tested but not linted.

`query.go` now skips captures whose name begins with `_`, which is the convention tree-sitter itself uses for internal captures. Without it, the query has to capture the whole `@moduledoc """..."""` and take the delimiters off with `Delims` — which for the single-quoted `@doc "..."` form would also strip any quote written inside the prose.

With it, the queries capture `quoted_content` (the body of the string) and the extracted text needs no delimiter stripping at all. `Delims` is just `#`.

This is a small behaviour change to shared code: any existing query using an `_`-prefixed capture name would stop having that capture linted. I could find none in-tree, and linting a capture that exists for a predicate looks like the unintended reading, but flag it in case you'd rather have it behind a language flag.

### Notes

- `github.com/smacker/go-tree-sitter/elixir` is already vendored, so this adds no dependency.
- `.heex` (Phoenix HTML templates) is deliberately left alone — no grammar for it in the set, and mapping it onto HTML would reduce it to `<!-- -->` comments.

### Tests

- `testdata/comments/{in,out}/8.ex`, `9.exs` — extraction fixtures covering heredoc docs, the single-quoted form, sigil docs (`~S"""`), a line-comment run, a trailing comment, and the prose-free attributes.
- `testdata/e2e/lint.yaml` — an `elixir` case against `testdata/fixtures/formats/test.ex`; all seven reported columns are exact.
- `TestPredicateOnlyCapture` and `TestDocAttributesWithoutProse` in `comments_test.go`.

`go test ./internal/...` is green.

Happy to split the `query.go` change into its own commit or PR if you'd prefer to take them separately.

### DIFF
--- a/internal/core/format.go
+++ b/internal/core/format.go
@@ -58,6 +58,7 @@ var FormatByExtension = map[string][]string{
 	`\.(?:css)$`:                             {".css", "code"},
 	`\.(?:cs|csx)$`:                          {".c", "code"},
 	`\.(?:dita)$`:                            {".dita", "markup"},
+	`\.(?:ex|exs)$`:                          {".ex", "code"},
 	`\.(?:go)$`:                              {".go", "code"},
 	`\.(?:hs)$`:                              {".hs", "code"},
 	`\.(?:html|htm|shtml|xhtml)$`:            {".html", "markup"},
--- a/internal/lint/code/comments_test.go
+++ b/internal/lint/code/comments_test.go
@@ -71,3 +71,50 @@ func TestComments(t *testing.T) {
 		}
 	}
 }
+
+// TestPredicateOnlyCapture covers the underscore convention: a capture that
+// exists for a predicate to test is not itself linted.
+//
+// Without it, testing one node and extracting another is impossible -- the
+// only testable node is the one you extract -- which is what the Elixir doc
+// attributes need (the attribute name decides, the heredoc is the prose).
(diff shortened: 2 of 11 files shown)


jdkato:
Looks good, thanks

PR #1158, opened 2026-08-28
feat: add versioned package installation and caching

httphypixelnet:
## Summary

Add version-aware package installation for packages hosted on GitHub. Package
references can now specify a Semver constraint, and Vale resolves the
latest matching GitHub release before installing it.

Add an on-disk cache for GitHub release metadata and downloaded release
archives. Cache locations follow platform conventions and can be overridden
with `VALE_CACHE`.

## Details

- Resolve `Packages` entries such as `Google@0.7.0` and version ranges against
  GitHub releases.
- Prefer matching release assets, then fall back to the release zipball.
- Cache release metadata with ETags and cache downloaded release archives.
- Add package end-to-end coverage for exact versions, version ranges, invalid
  versions, and cache output.
- Modify E2E suite to work with the updated methods

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -45,7 +45,7 @@ The end-to-end suite shells out to external converters, so these also need to be
 * [rst2html](https://docutils.sourceforge.io/docs/user/tools.html#rst2html-py), installed with [docutils](https://pypi.org/project/docutils/) or [Sphinx](https://www.sphinx-doc.org/)
 * [xsltproc](http://xmlsoft.org/xslt/xsltproc.html)
 * [dita](https://www.dita-ot.org/download) (v3.6+)
-* [mdx2vast](https://www.npmjs.com/package/mdx2vast) (`npm install -g mdx2vast`)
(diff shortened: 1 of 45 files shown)


jdkato:
Thanks for the work here, but I think this is out of scope for Vale.

For pinning, this is already supported: give a release URL in place of the name.

```ini
Packages = https://github.com/errata-ai/Google/releases/download/v0.7.0/Google.zip
```

That works for any host, and `BasedOnStyles` reads the same either way ([docs](https://docs.vale.sh/keys/packages#pinning-a-version)).

What's left beyond that is resolving version *ranges*, and that's where I'd rather Vale not go:

- Ranges need a semver resolver, a GitHub API client, and a cache to keep its traffic tolerable -- GitHub-only, where `Packages` is host-agnostic today.
- They'd also need a lockfile to be worth trusting. Without one, `@^0.7` re-resolves on every `sync`, so it floats with a ceiling rather than pinning -- not what users expect.

The middle ground you're after is a real want. It's just the expensive end, and I'd rather keep Vale at the two ends than maintain a package manager to reach the middle.

Issue #1124, opened 2026-07-30
`sequence` rules never fire outside paragraphs: the declared `scope` is silently replaced with `sentence`


wasaga:
### Environment

- **Vale version**: 3.16.0 (also reproduced by building `391e3c0b` from source)
- **Operating system**: macOS 26.5.1, arm64
- **Installation method**: Homebrew (`vale 3.16.0`); plus `go build ./cmd/vale` at `391e3c0b` for the source comparison

### Summary

A `sequence` rule matches a sentence in a paragraph. It silently misses the **identical** sentence in a list item, heading, table cell or blockquote. Vale appears to discard whatever `scope` the author writes, without an error or a warning.

`sequence` is the only extension point that can reach part-of-speech data. So POS-based rules cannot reach list items at all — which, in procedural documentation, is where most of the prose lives.

### Expected vs. actual

The rule below matches four adjacent nominal tokens. It is an arbitrary probe, chosen only because it needs part-of-speech data and fires reliably. The same sentence appears in each block type.

"Expected" is what an equivalent `existence` rule does in the same run — the closest thing to a baseline:

| Block | `existence` (baseline) | `sequence`, no scope | `sequence`, `scope: [list, paragraph, heading]` |
| --- | --- | --- | --- |
| paragraph | match | match | match |
| ordered list item | match | **no match** | **no match** |
| unordered list item | match | **no match** | **no match** |
| heading | match | **no match** | **no match** |
| blockquote | match | **no match** | n/a — not requested |
| table cell | match | **no match** | n/a — not requested |

So the content is reaching Vale; it just isn't reaching `sequence`. And the third column is what I'd most like to flag: declaring a scope changes nothing, so the two `sequence` columns are identical.

I expected one of two things: that Vale would honour `scope` on a `sequence` rule, or — if `sequence` genuinely cannot support those scopes — that writing one would error rather than silently do nothing.

### Minimal reproduction

`.vale.ini`

```ini
StylesPath = styles
MinAlertLevel = suggestion

[*.md]
BasedOnStyles = T
```

`styles/T/S.yml`

```yaml
extends: sequence
message: "Sequence matched '%s'."
level: error
tokens:
  - upos: NOUN|ADJ
  - upos: NOUN|ADJ
  - upos: NOUN|ADJ
  - upos: NOUN|ADJ
```

`s.md`

```markdown
Update the default database connection pool size limit.

1. Update the default database connection pool size limit.
```

```console
$ vale --output=line s.md
[...]

Issue #899, opened 2024-09-20
Allow for counting multiple elements of the same kind in sequence rules


aireilly:
### Check for existing issues

- [X] Completed

### Describe the feature

I am attempting to write a Vale rule to highlight ambiguous pronouns in a sentence. I want the rule to fire on a pronoun (PRP) or personal pronoun possessive (PRP$) only when it is preceded by 2 or more nouns in the sentence. I don't think this is possible currently. See https://studio.vale.sh/s/fa39e2d796458f6e8903cb5728807282

Note the last line in the fixture. The sequence rule doesn't handle multiple instances of the same POS tag.

Something like a min or max count field in the sequence rule would be useful for rules like this I think.

```yaml
---
extends: sequence
message: "Avoid ambiguous pronouns"
level: warning
ignorecase: true
tokens:
  - tag: NN|NNP|NNPS|NNS|PERSON|GPE
    skip: 8
    min: 2
  - pattern: \w+
    tag: PRP|PRP$|WP|WP$|SYM
```

michaeltlombardi:
This would also be useful in cases where I need to support one-or-more tokens of a given tag. As far as I can tell, this use case is distinct from defining `skip`.

jdkato:
Implemented for the next release — and your proposed syntax is exactly what shipped:

```yaml
extends: sequence
message: "Avoid ambiguous pronouns."
level: warning
tokens:
  - tag: NN|NNP|NNPS|NNS
    skip: 8
    min: 2
  - pattern: \w+
    tag: PRP|PRP$|WP|WP$
```

`min: N` repeats the token's whole group, `skip` window included, so the rule above reads "a noun within eight words, then another noun within eight words, then the pronoun" — occurrences can be scattered, they don't need to be adjacent. With no `skip`, `min` means consecutive occurrences (`tag: JJ, min: 2` = two adjectives in a row), which covers @michaeltlombardi's case for any bounded count. The default is 1, so existing rules are unchanged.

PR #1162, opened 2026-08-31
feat: add max/min count thresholds to sequence rules

### LINKED ISSUE #1161: sequence has no way to count occurrences of the whole pattern within a scope, only occurrence does, and occurrence has no tag support
## Problem

`sequence` matches a tagged multi-token pattern but alerts once per match. It has no threshold or count concept. `occurrence` has a count threshold (`Max`/`Min`) but only matches a raw regex, no `tag`/`upos` field at all. Neither can count occurrences of a *tagged* pattern within a scope.

## Motivating case

`tbhb/vale-ai-tells` includes `VerbTricolonDensity.yml`, an `occurrence` rule (`max: 1`) meant to flag more than one AI-cliché verb tricolon per paragraph. Its regex can't require the matched words be verbs, so it fires on plain noun-phrase lists too. The single-instance version of the same rule has an unmerged `sequence`-based fix using real POS tags. The density version can't get the same fix today: no way to express "this tagged pattern needs to occur 2+ times in one paragraph."

## Root cause

`NewSequence` unconditionally rewrites every declared scope to sentence-level via `sentenceScope()`. `Run()` is invoked once per sentence regardless of what a rule declares. `scope: paragraph` on a `sequence` rule currently has no effect, it collapses to the same "every sentence, everywhere" behavior as an undeclared scope.

Practical effect: a paragraph with two real matches split across two sentences doesn't produce any density alerts, since each sentence individually has only one match, at or under any reasonable threshold.

## Proposed fix

Contained to `sequence.go`, plus small additions to the NLP infrastructure it depends on:

- When a rule opts into count-threshold behavior, skip the `sentenceScope()` rewrite. `Run()` is then called once per the rule's actually-declared scope (e.g. once per paragraph).
- Tag each sentence of that scope separately, instead of tagging the whole block once and inferring sentence boundaries afterward. The existing tagging function already segments and tags sentence by sentence internally, so this reuses information the codebase already had.
- Sentence membership becomes a direct fact this way, since it's determined by which tagging call produced a token, so a match can never span two sentences by construction and the match-walk doesn't need a separate boundary guard.
- Threshold/message logic mirrors `occurrence`'s existing `Max`/`Min` convention. It runs over a correctly-scoped match set built one sentence at a time.

<details>
<summary>Alternative considered: extending `occurrence.go`</summary>

`occurrence` already runs once per a rule's real declared scope correctly, including working `scope: raw`, and could reuse `sequence`'s tag-matching helpers directly (same package).

Set aside: `occurrence` runs concurrently across rules matching one block; `sequence` is deliberately excluded from concurrent execution because its tagging cache isn't synchronized (confirmed with Go's race detector, not just by reading the code). Fixable, but it touches a second file plus shared dispatch infrastructure other check types rely on, and gives `Occurrence` two structurally different, mutually exclusive matching modes. The `sequence.go`-only fix touches less shared code.

</details>

## Related work

- `min: N` per-token repetition, #899 (closed 2026-08-26)
- `sequence` `scope: paragraph` fix, #1124/#1126 (closed 2026-07-30)

This request is the remaining piece: counting the whole pattern within its real declared scope, not one token within it.

## Status

Not implemented. #1162 (an earlier full implementation) was closed, its scope was judged too large relative to the motivating case, and it leaned on `NLPEndpoint`, which is out of scope. Open to discussing a narrower approach.

theredspoon:
Closes #1161

**Review note:** this branch is rebased on #1169 (needed underneath for correctness, see Fix). Only the commit `feat: add max/min count thresholds to sequence rules` belongs to this PR. Click into it on the **Commits** tab to see this PR's actual diff. The **Files changed** tab includes #1169's content too, and will shrink once it merges into `v3` and this rebases.

## Problem

`sequence` matches tagged patterns but only alerts once per match. It has no count threshold. `occurrence` has `Max`/`Min` but only matches a raw regex, no `tag`/`upos`. Root cause: `NewSequence` unconditionally narrowed every declared scope to sentence-level, so a paragraph with two real matches split across two sentences stayed silent, since each sentence individually had only one match.

`tbhb/vale-ai-tells`'s `VerbTricolonDensity.yml` needs exactly this. Its `occurrence` regex can't require the matched words be verbs. A `sequence` rule can, and now finally gets a count threshold too.

## Fix

`Run` now tags each sentence of a rule's real declared scope separately (reusing `internal/nlp/prose.go`'s existing per-sentence tagging), instead of tagging the whole block once and inferring boundaries afterward. Sentence membership becomes a direct fact, so a match can't span two sentences by construction.

- Threshold rules (`Max`/`Min` set) keep their real declared scope instead of being narrowed to sentence-level.
- An undeclared scope on a threshold rule defaults to paragraph plus prose-container scopes, matching plain-rule behavior, via one shared list instead of two hand-copied ones.
- `File.Sentences` routes through the same local/remote segmentation dispatch the rest of the codebase already uses.

<details>
<summary>Why the redesign, not a smaller patch (root cause detail)</summary>

Sentence boundaries used to be inferred after tagging a whole block once: compare each word's offset against each sentence's own offset. That broke down for a remote endpoint's unpositioned tokens, which carry no offsets to compare. Tagging each sentence separately makes sentence membership a direct fact instead, determined by which loop iteration produced the word.

</details>

## Performance

A real but small per-rule cost for local English tagging, 6-8% (see below for the benchmark). A non-English remote endpoint pays up to 2 round-trips per sentence instead of 1; match-walk cost drops from O(N²) per block to a sum of smaller O(N²/k) terms across k sentences.

<details>
<summary>Full performance reasoning</summary>

For a style's `sequence` rules as a whole, this is a small, consistent cost for local English tagging, not the neutral-to-noise result originally claimed here. The new per-sentence segmentation pass is cached per file, so only the first rule to touch a sentence pays for the actual segmentation work; every other rule sharing that file's cache gets a map-lookup hit instead. That part of the claim holds. What doesn't: `BenchmarkSequencePlainRuleSentencesOverhead` (`internal/check/sequence_bench_test.go`), compared with `benchstat` against a control that calls the same matching logic without ever going through `File.Sentences`, measures one plain rule against a cold cache at 8.1% slower (16.49µs vs a 15.26µs baseline, p<0.001, n=10). Five plain rules sharing one file's cache still cost 5.7% more per rule on average (5.10µs vs 4.83µs baseline, p=0.02, n=10), not the noise-level difference this section originally claimed. The remainder isn't the segmentation call itself. Cache hits there cost about 19ns (see the next benchmark). It's the fixed per-call cost of `Run`'s new sentence-loop wrapping, which every rule pays whether or not its cache lookup was a hit.

`sequenceMatches`' match-walk is quadratic in the number of words per call. Splitting a block into sentences turns one large quadratic term into a sum of smaller ones, `N²/k` instead of `N²` for `k` sentences. Each sentence also pays a small, fixed per-call cost: tokenization and a fresh lookup map. A block of many very short sentences can let that fixed cost outweigh the quadratic saving. Realistic prose doesn't hit this.

Sentence segmentation is cached: a repeated call against the same text hits a map lookup. `BenchmarkFileSentencesCache` (`internal/check/sequence_bench_test.go`), compared with `benchstat`, measures a cold segmentation at 4.93µs against a cached lookup at 19.3ns: a 99.6% reduction, roughly 256x faster (p<0.001, n=10), against a three-sentence fixture. That is well past the ~140x this section originally estimated before a real benchmark existed. The exact multiple depends on the fixture: cache-hit cost is a fixed map lookup, but cold cost scales with how much there is to segment.

A non-English document behind a remote NLP endpoint pays an extra HTTP round-trip, but not a flat doubling. A plain (sentence-scoped) rule now makes one segmentation call and one tagging call per sentence, two round-trips where there was one before. A `max`/`min` rule segments its whole scope in one call, then tags each sentence individually: one extra segmentation round-trip per block rather than per sentence, shrinking as a fraction of the total as a block's sentence count grows. Both calls are cached per file, so a style with many `sequence` rules pays this once per unique sentence, not once per rule. Local English tagging, the common case, never touches this path at all.

</details>

## Testing

- cross-sentence aggregation, Max/Min combined
- cross-sentence boundary guard under both local and remote tagging, including a remote tagger normalizing token text
- undeclared-scope threshold rule dispatching into headings and list items
- `internal/e2e` case (`checks/sequence/max`) for the density-across-sentences motivating case, end-to-end through a real `.vale.ini`/style/document
- `TestSequenceActionResolvesSuggestion`: a threshold (`max`/`min`) rule with a configured `action` (`replace`/`remove`/`suggest`) now resolves and carries its suggestion, the same as a plain rule already does. This was a real gap the latest rebase surfaced, not part of the original change: extracting the per-sentence match loop into its own `matchesIn` helper predates `resolveFix` existing on this codepath at all, so the extraction never called it. Confirmed red without the call, green with it restored.

Full repo suite and `-race` are clean, including `internal/e2e`. Confirmed no reintroduction of the concurrency risk `sequence` was historically excluded from concurrent dispatch for.

<details>
<summary>Alternative considered: extending occurrence.go instead (rejected)</summary>

`occurrence` could reuse `sequence`'s matching helpers, but it runs concurrently across rules while `sequence`'s tagging cache isn't synchronized. Extending it would touch shared dispatch infrastructure and give `Occurrence` two structurally different matching modes.

</details>

## Related

`min: N` per-token repetition (#899) · `scope: paragraph` fix (#1124/#1126) · `sentenceScope` negated-scope double-report fix, split out to #1169 · redundant round-trip elimination for plain rules, explored separately in #1170, closed since it only makes sense as a commit here, not standalone

### DIFF
--- a/internal/check/sequence.go
+++ b/internal/check/sequence.go
@@ -611,11 +611,15 @@ func sentenceScope(declared []string) []string {
 			scopes = append(scopes, s)
 			continue
 		}
-		// Negation applies to the part being excluded, so it stays in front:
-		// `~list` narrows to sentences outside a list, not to something
-		// outside `sentence.list`.
-		if neg, found := strings.CutPrefix(s, "~"); found {
-			scopes = append(scopes, "~"+neg)
+		// A bare negated term names only what to exclude and never mentions
+		// `sentence` itself, so asksForSentence (scope.go) skips every
+		// `sentence.*` fragment block for it and the rule matched the whole
+		// unsegmented block instead -- narrowing to `~list` alone was a
+		// no-op. AND-ing `sentence` in front keeps the exclusion and still
+		// narrows: `sentence&~list` is "sentences outside a list", which is
+		// what the rule actually needs.
+		if strings.HasPrefix(s, "~") {
+			scopes = append(scopes, "sentence&"+s)
 			continue
 		}
 		// `paragraph` names no block of its own. Splitting wraps every block
--- a/internal/check/sequence.go
+++ b/internal/check/sequence.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/jdkato/prose/v3/segment"
 	"github.com/jdkato/prose/v3/tag"
 	"github.com/mitchellh/mapstructure"
 	rx "github.com/vale-cli/vale/v3/internal/regex"
@@ -89,6 +90,43 @@ type Sequence struct {
 	//
 	// Empty means prose's own tagger, which is what every existing rule gets.
 	Model string
+
+	// `max`/`min` (`int`): A count threshold on how many times the whole
+	// token sequence, not one token within it, occurs in the scope.
+	//
+	// Same names and meaning as `occurrence`'s: unset (the zero value) keeps
+	// the existing behavior of one alert per match. Setting either turns the
+	// rule into a density check, reporting a single alert with the match
+	// count in its message instead of one alert per match. `min: 2` on a
+	// single token already covers that token's own repetition; this is for
+	// repetition of the whole pattern.
+	//
+	// A plain `sequence` rule's declared scope is always narrowed to
+	// sentences (see sentenceScope), since it reads part-of-speech data
+	// that is tagged a sentence at a time. Setting `max` or `min` skips
+	// that narrowing instead: Run is then called once per the rule's real
+	// declared scope -- once per paragraph for `scope: paragraph` -- so
+	// this counts matches across every sentence of that scope, not just
+	// one. Run itself tags and walks each of that scope's sentences one at
+	// a time (see Run), so a single match can never span two of them: "two
+	// tricolons in two different sentences of the same paragraph" trips
+	// `max: 1` the same as two tricolons in one sentence would.
+	//
+	// An undeclared scope (see NewSequence) reaches a real paragraph plus
+	// every other prose container -- a heading, list item, blockquote,
+	// table cell/header/caption, or figure caption (core.ProseContainerScopes)
+	// -- but, deliberately, nothing narrower than that: a frontmatter key
+	// (`text.frontmatter.<key>`), a code comment (`text.comment.block` /
+	// `text.comment.line`), or link/image alt text (`text.attr.alt`) are
+	// each prose a plain (sentence-scoped) rule's own undeclared scope does
+	// reach, but a `max`/`min` rule's does not. Aggregating a count needs a
+	// real, named container to aggregate over -- unlike a plain rule, which
+	// just wants "every sentence, everywhere" -- and these three are small,
+	// narrow fragments a style is unlikely to want density-checked on their
+	// own; naming one explicitly in `scope` still reaches it like any other
+	// declared scope would.
+	Max int
+	Min int
 
 	// `exceptions` (`[]string`): Regexes matched against the sentence; a
 	// sequence match that *begins inside* one of their regions is dropped.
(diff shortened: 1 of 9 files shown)


theredspoon:
recheck

theredspoon:
I have read the CLA and I agree to it.

jdkato:
I'm closing this. One rule in one third-party style shouldn't drive a rewrite of how every `sequence` rule runs, plus new user-facing semantics (`max`/`min` on `sequence`), with no discussion.

Two specifics:

- It builds on `NLPEndpoint`, which isn't documented and never left WIP. This adds a new code path, a cache, and a dozen tests hardening one, with a network call inside `sequence.Run`.
- The scope. The feature is a threshold; the change is 2,000 lines re-tagging every existing rule per sentence, plus a benchmark suite.

We can discuss what something like `VerbTricolonDensity.yml` might require, but I take API-related changes seriously because its the core entry point into Vale.

PR #1166, opened 2026-09-02
feat: expose per-check alert counts to metric formulas

### LINKED ISSUE #1163: Expose per-check alert counts to metric formulas, for composite/density scoring across multiple rules
## Problem

`metric` rules already support arbitrary custom formulas (Tengo expressions evaluated against `f.ComputeMetrics()`), not just the built-in readability formulas. What's missing is an input: there's no way for a `metric` formula to know how many times other rules have already fired, so a rule combining signals from several other rules is not expressible.

## Motivating case

A style package like `tbhb/vale-ai-tells` includes about 110 independent pattern rules, each an isolated per-instance match, none currently setting `limit:`. One or two firing in a document reads as normal writing. Many different ones firing across the same document is a stronger signal. This would make that signal expressible as an ordinary `metric` rule, without requiring every contributing rule to be reconfigured first.

## What exists today, and why it doesn't cover this

`File.AddAlert` has a `limits map[string]int` keyed by check name, but it's tied to a different, opt-in feature: it only increments when a rule sets `limit: N` in its own YAML, to cap how many alerts that rule reports. Checked all 110 rules in `tbhb/vale-ai-tells`, including the three in the example formula below: none set `limit:`. For a package like that, `f.limits` stays at zero for every rule, it doesn't give a `metric` formula what it needs.

## Proposed fix

A new, unconditional per-check counter, incremented on every reported alert regardless of `Limit`, separate from the existing opt-in reporting-cap mechanism.

Exposed to `metric` formulas as an indexable object rather than as individual identifiers, since check names come from user-chosen style and rule filenames and can contain arbitrary characters that don't map cleanly onto Tengo identifiers:

```
check["AITells.FigurativeOwns"] + check["AITells.HedgingPhrases"] + check["AITells.EmptyPadding"] > 3
```

A check that's loaded and could apply to the current file, but hasn't fired yet, reads as `0`. A name that doesn't correspond to a real, applicable check produces a clear error naming it, rather than silently reading as zero.

An unscoped `metric` rule runs at `scope: summary` by default, so the composite score this enables is a whole-document signal.

<details>
<summary>Why dispatch order is safe for this</summary>

Vale's lint pipeline runs in two phases. Every format (`lintMarkdown`, `lintADoc`, `lintRST`, `lintDITA`, `lintOrg`, `lintXML`, `lintHTML`) funnels into `lintHTMLTokens`, which walks the whole document running every block-scoped rule (`existence`, `sequence`, `occurrence`, etc.), each synchronously reporting through `f.AddAlert`. Only after that walk fully completes does `lintSizedScopes` build the `scope: summary` block and run unscoped `metric` rules. Any per-check counter populated in `AddAlert` is complete by the time an unscoped, `summary`-scoped `metric` formula reads it. A `metric` rule can also declare a narrower scope (`paragraph`, `heading`, and so on), which runs mid-walk instead, before the document finishes, so it sees counts as of that point, not the final total. The examples in this issue are all whole-document rollups and are meant to be unscoped.

</details>

## Other formulas this enables

A few beyond the `vale-ai-tells` density case above:

- **Readability consensus.** Pick one formula per distinct measurement instead of summing all seven built-in ones (several are near-duplicates of each other):

  ```
  check["Readability.FleschKincaid"] + check["Readability.ColemanLiau"] + check["Readability.GunningFog"] + check["Readability.SMOG"] > 2
  ```

  `> 2` requires a real majority, at least three of four, instead of nagging the moment one formula crosses its own line.

  <details>
  <summary>Why these four, and other notes on this example</summary>

  Several of the seven built-in formulas are near-duplicates measuring the same underlying stats (`FleschReadingEase` is `FleschKincaid`'s two inputs rescaled; `AutomatedReadability` and `LIX` both reuse `ColemanLiau`'s character-length approach), so a naive sum of all seven mostly double-counts three or four real signals wearing seven labels. Each formula picked here is built from a different input: `FleschKincaid` from average syllables per word, `GunningFog` from percent of complex words, `ColemanLiau` from average word length in characters, and `SMOG` from polysyllabic-word density, the only one of the four that skips sentence length entirely.

  Combining readability formulas wasn't previously impossible: the built-in `readability` check type (`internal/check/readability.go`, separate from the `metric`-based `Readability` style used above) already lets a rule pick several metrics and average their raw grade values against one threshold. That's a different approach from voting. Averaging blends every input into one number, so one formula reading unusually high can still pull the average past the threshold even if the other four read comfortably below it. Voting resists that: it only fires once an actual majority agrees. Both are useful. This example is specifically about the vote.

  `check[...]` is always a real per-document count. It reads as 0 or 1 here specifically because these four `Readability.*` rules are unscoped, which runs them at `scope: summary` and reports at most one alert per document, so summing four of them really is counting votes. That's not true of every check: `check["Bugs.MatchCase"]` in the consistency-scoring example below is a multi-value count, since an `existence`/`substitution` rule can match several times in one document.

  </details>

- **Consistency scoring.** Several narrow terminology/case-style rules (`Bugs.MatchCase`, `Bugs.TermCase`, `Bugs.SameCase`, and similar) each catch one kind of inconsistency. One hit is a normal typo-level nit. Several different ones firing across the same document signals a process breakdown, maybe no shared glossary, maybe several authors who never synced on terminology, maybe a style guide nobody enforced. Summing their counts turns that into a document-level signal one rule can't express alone.
- **Weak-writing composite.** Passive voice, weasel words, and wordiness are each a mild, individually-ignorable nudge. Combined past a threshold, they're a real "tighten this section" signal one rule alone can't express.
- **Accessibility risk rollup.** Several narrow a11y pattern rules (missing alt-text conventions, heading-level jumps, and the like) combined into one composite flag when enough of them fire together in the same document.

None of these need new checks, just a `metric` rule combining counts from checks that already exist.

## Status

Implemented in #1166.

theredspoon:
Closes #1163

## Problem

`metric` rules already support arbitrary custom Tengo formulas, not just built-in readability formulas. What's missing is an input: there's no way for a `metric` formula to know how many times other checks have already fired, so a rule combining signals from several other rules isn't expressible.

## Motivating case

A style package like `tbhb/vale-ai-tells` includes about 110 independent pattern rules, each an isolated per-instance match. One or two firing in a document reads as normal writing. Many different ones firing across the same document is a stronger signal. This makes that signal expressible as an ordinary `metric` rule, so `check["AITells.FigurativeOwns"] + check["AITells.HedgingPhrases"] + check["AITells.EmptyPadding"] > 3` becomes possible.

## Fix

Contained to `internal/core/file.go`, `internal/check/metric.go`, a new `internal/check/check_counts.go`, `internal/lint/lint.go`, and a small simplification in `cmd/vale/command.go`'s `ls-metrics` command:

- `AddAlert` increments a per-check counter in a dedicated `File.checkCounts` field every time an alert is actually appended, so it only counts alerts that were actually reported, not deduped or hidden ones. This field is decoupled from `f.Metrics`, which document content can influence via HTML tag-name bookkeeping, so nothing in the document itself can forge a count for a real check.
- That counter is exposed to `metric` formulas as an indexable object: `check["Style.Rule"] + check["Style.OtherRule"] > 3`.
  - A check that's loaded and applicable to the current file reads as `0` until it actually fires.
  - A name that doesn't correspond to a real, applicable check produces a clear error naming it, so a typo in a formula raises an error instead of silently evaluating as if the check simply hadn't fired.
- Which checks are known and applicable is computed per file, respecting the same extension, section, and style-toggling rules that already decide whether a check runs at all, not just the full set of everything loaded anywhere in the config.
- Existing opt-in `limit:`/`f.limits` behavior is unaffected. Both mechanisms coexist, gated through the same `f.Alerts` append.

Counters are exposed as an indexable object (`check["Style.Rule"]`) rather than as individually sanitized identifiers (`check_Style_Rule`), since check names come from user-chosen style and rule filenames and can contain arbitrary characters.

<details>
<summary>Why indexing instead of sanitized identifiers</summary>

Flattening check names into identifiers risked two differently-named checks colliding on the same identifier, and made it hard to know which identifiers a formula actually references without re-implementing Tengo's own scoping rules. An indexable object has neither problem: no flattening step to collide on, and a lookup checks the real, applicable check names directly.

</details>

## Performance

`AddAlert`'s counter increment is one `O(1)` map write. `check[...]` lookups each resolve to one map read, with no parsing.

`LoadedChecks` construction takes about 4.6µs against a 1.46ms full lint pass, roughly 0.3% of total time (see below for the benchmark). `ComputeMetrics` makes one pass over `f.Metrics`.

<details>
<summary>Benchmark methodology, and a correction to an earlier estimate</summary>

`LoadedChecks` is built once per file: a loop over every loaded check name, with a couple of map lookups and one short scan over `f.BaseStyles` per check. Benchmarked with `BenchmarkLoadedChecksConstruction` and `BenchmarkLintSyntheticAITells` (`internal/lint/loaded_checks_bench_test.go`) against a synthetic 110-rule `existence` style shaped like the motivating case, linting a realistic ~5KB document: across 10 runs (`go test -bench=. -benchtime=2s -count=10`, summarized with `benchstat`, on an Apple M3 Pro/darwin-arm64), `LoadedChecks` construction takes 4.60µs ± 6%, against 1.46ms ± 1% for the full lint pass. An earlier, uncommitted ad-hoc measurement had put the full pass at ~550µs and the share at ~0.7%; those numbers didn't reproduce, and are replaced here by this committed, reproducible benchmark.

</details>

## Testing

Coverage includes:

- the counter only incrementing for alerts that were actually reported
- a real multi-rule `metric` formula combining counts from two checks
- a missing-but-applicable check reading as zero, including under a lowered `MinAlertLevel` and via the `--minAlertLevel` CLI flag
- an unknown check name producing a real error
- two check names differing only in punctuation resolving correctly and independently
- a regression test for the crafted-tag-injection case described above
- an `internal/e2e` case (`checks/metric/check-counts`) exercising `check[...]` end to end through a real `.vale.ini`/style/document, verified against the pre-feature commit to confirm it fails there with a real Tengo compile error
- a `scope: paragraph` `metric` rule reading `check[...]` at three different points in a document, confirming it sees only the check count recorded so far at its position in the walk, not the document's eventual final total, unlike the default `scope: summary` case covered elsewhere in this list
- a regression test (`TestNestedRuleDirectoryDisableRegression`) for the `checkApplies`/`RuleForAlert` reconciliation needed once a rule can be loaded from a nested style directory: a rule name with more than one dot (e.g. `Std.dates.TimeFormat`) must resolve through the real, loaded rule names rather than being truncated to its first two dot-separated segments, or a section override disabling it by its exact name would be silently missed

### DIFF
--- a/cmd/vale/command.go
+++ b/cmd/vale/command.go
@@ -248,7 +248,16 @@ func printMetrics(args []string, _ *core.CLIFlags) error {
 			"'%s' contains no lintable files", args[0]))
 	}
 
-	computed, _ := linted[0].ComputeMetrics()
+	return printMetricsResult(linted[0])
+}
+
+// printMetricsResult reports f's computed metrics as JSON.
+func printMetricsResult(f *core.File) error {
+	computed, err := f.ComputeMetrics()
+	if err != nil {
+		return err
+	}
+
 	return printJSON(computed)
 }
 
--- a/cmd/vale/metrics_test.go
+++ b/cmd/vale/metrics_test.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/vale-cli/vale/v3/internal/core"
+)
+
+// TestPrintMetricsResultSucceedsWithoutCollision pins the ordinary,
+// non-colliding path: printMetricsResult must still succeed and print
+// normally when ComputeMetrics finds nothing ambiguous.
+func TestPrintMetricsResultSucceedsWithoutCollision(t *testing.T) {
+	cfg, err := core.NewConfig(&core.CLIFlags{IgnoreGlobal: true})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	f, err := core.NewFile("A document with real prose for a word count.", cfg)
+	if err != nil {
+		t.Fatal(err)
+	}
+	f.Summary.WriteString("A document with real prose for a word count.")
+
+	if err = printMetricsResult(f); err != nil {
+		t.Fatalf("expected no error, got: %v", err)
+	}
+}
+
+// TestPrintMetricsResultNoLongerErrorsOnSanitizedKeyCollision pins the
+// metric-check-counts redesign's effect on this CLI surface: with check
+// names no longer flattened into Tengo identifiers at all, there is no more
+// sanitized-key collision for ComputeMetrics to detect or for
+// printMetricsResult to propagate specially -- see item 4 of the redesign
+// (cmd/vale/command.go's printMetricsResult should "revert to something
+// much simpler, there's no more collision to propagate specially").
+//
+// This reproduces the f.Metrics = {"words": 999} setup the deleted
+// TestPrintMetricsResultSurfacesCollision (an old-mechanism test pinning
+// the now-removed collision machinery) used to assert an error for, but
+// asserts the opposite outcome: printMetricsResult must now succeed.
+func TestPrintMetricsResultNoLongerErrorsOnSanitizedKeyCollision(t *testing.T) {
+	cfg, err := core.NewConfig(&core.CLIFlags{IgnoreGlobal: true})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	f, err := core.NewFile("A document with real prose for a word count.", cfg)
+	if err != nil {
+		t.Fatal(err)
+	}
+	f.Summary.WriteString("A document with real prose for a word count.")
+	f.Metrics["words"] = 999
+
+	if err = printMetricsResult(f); err != nil {
+		t.Fatalf("expected no error under the new design (there is no more "+
+			"collision-detection machinery left to trip), got: %v", err)
+	}
+}
--- a/internal/check/check_counts.go
+++ b/internal/check/check_counts.go
@@ -1,0 +1,78 @@
+package check
+
+import (
+	"fmt"
+
+	"github.com/d5/tengo/v2"
+)
+
+// checkCounts is a Tengo object exposing every loaded check's per-document
+// alert count by its real, unflattened name (e.g. "Style.Rule"), added to a
+// `metric` formula's parameters under the "check" key. A formula indexes it
+// directly:
+//
+//	check["AITells.FigurativeOwns"] + check["AITells.HedgingPhrases"] > 3
+//
+// This replaces the old design, where a check name was flattened into a
+// sanitized Tengo identifier (check_Style_Rule) for direct reference in a
+// formula -- a step where two distinct names could sanitize to the same
+// identifier, with no way to tell them apart afterward. Indexing by the
+// real name makes that collision structurally impossible: there's no
+// flattening step left to collide on.
+//
+// It also fixes a silent-typo problem the old design had no way to catch:
+// IndexGet distinguishes a check that's genuinely loaded but simply never
+// fired on this document (reads as 0) from a name that was never loaded at
+// all -- almost always a typo in the formula -- which returns a real error
+// instead of silently reading 0 either way.
+type checkCounts struct {
+	tengo.ObjectImpl
+	// counts holds the raw, unsanitized per-check alert count for this
+	// document, keyed by the check's real name (e.g. "AITells.FigurativeOwns"
+	// -> 3). A check absent from counts simply never fired.
+	counts map[string]int
+	// known holds the set of check names actually loaded for this run --
+	// see core.File.LoadedChecks -- which is what lets IndexGet tell a
+	// never-fired check apart from one that doesn't exist.
+	known map[string]bool
+}
+
+// newCheckCounts builds a checkCounts object from counts (raw per-check
+// alert counts) and known (the set of check names loaded for this run).
+func newCheckCounts(counts map[string]int, known map[string]bool) *checkCounts {
+	return &checkCounts{counts: counts, known: known}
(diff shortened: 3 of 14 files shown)


theredspoon:
recheck

theredspoon:

jdkato:
Thanks for the work here, and for the thorough write-up. I'm going to close this rather than iterate on it.

- **A rule's output stops depending on the document alone.** With `check[...]`, a `metric` result changes with which other rules are enabled, at what level, in which section, and what survived deduplication and `limit`. The same text under a slightly different `.vale.ini` gives a different number, and the alert can't say why.
- **Unknown names are fatal, and "unknown" is decided per file.** A formula that names a check one section disables (`T.WordA = NO` under `[*.rst]`) fails every RST file and stops the run. That's ordinary configuration, not a typo.
- **Counts are read mid-walk for a scoped rule.** Add `scope:` to a working formula and it silently starts reporting "alerts before this paragraph".
- **Core code is reshaped for one rule type.** Splitting `shouldRun`, a per-file pass over every loaded rule, two new exported maps on `File`, and a second counter in `AddAlert` all exist to support the unknown-name error above.

The want itself is fair: many weak signals in one document add up. But that's a question about a run's output, not the text, and the JSON output already carries every alert with its check name, so a script or CI step can compute the composite today.

Closing #1163 with it.

Issue #356, opened 2021-07-08
[WIP] Multilingual, spaCy-powered NLP


jdkato:
All of the required pieces are finally in place to offer integration with [spaCy](https://spacy.io/):

> spaCy is a library for advanced Natural Language Processing in Python and Cython. It's built on the very latest research, and was designed from day one to be used in real products.

This allows Vale to support (1) rules written for any of spaCy's [supported languages](https://spacy.io/usage/models#languages) and (2) highly accurate (custom-trained, even) NLP.

If implemented well, I think this has the potential to easily 2x Vale's usefulness.

## Getting Started

The current version of Vale (v2.10.4) has unofficial support (since the implementation details are still a WIP) for this integration.

To get started, you'll need Vale (v2.10.4), Python 3.9, and [Pipenv](https://pipenv.pypa.io/en/latest/) installed. Next, follow the steps below:

1. Start the [`spacy-vale`](https://github.com/errata-ai/spacy-vale) API locally.

2. Create a `.vale.ini` file:

    ```ini
    StylesPath = styles
    MinAlertLevel = suggestion
    
    # This is the API started in step 1. 
    #
    # You'll need to change this value to the URL provided in uvcorn's output.
    NLPEndpoint = http://0.0.0.0:5000

    [*.md]
    # This is the language of the documents matched by the above glob pattern 
    # (`*.md`, in this case).
    Lang = en

    ...
    ``` 

3. Create a style/rules (see next section).

## Creating, testing, and debugging rules

The main entry point for NLP-based rules will be the [`sequence`](https://docs.errata.ai/vale/styles#sequence-v230) extension point. For example, an implementation of LanguageTool's [`WOULD_BE_JJ_VB`](https://community.languagetool.org/rule/show/WOULD_BE_JJ_VB?lang=en&subId=1) rule:

```yml
---
extends: sequence
message: "The infinitive '%[4]s' after 'be' requires 'to'. Did you mean '%[2]s %[3]s *to* %[4]s'?"
tokens:
  - tag: MD
  - pattern: be
  - tag: JJ
  - tag: VB|VBN
```

To help the testing process, you can use Vale Studio's _View Tags_ feature, which currently supports Markdown content written in Chinese, English, German, Russian, or Spanish.

![Screen Shot 2021-07-07 at 6 58 04 PM](https://user-images.githubusercontent.com/8785025/124850321-46532380-df55-11eb-832e-0a19c152d942.png)

[...]

PR #1167, opened 2026-09-02
fix: remote NLP endpoint failures surface as errors instead of panicking

theredspoon:
## Problem

`nlp.TextToTokens` (reached from `core.TextToContext`, reached from the `tag` CLI command) and `nlp.Info.Compute` (which runs during block construction, ahead of every rule's own `Run`) both panicked when a configured remote NLP endpoint's request failed: a network error, a non-2xx status, or a malformed response. That crashed the whole `vale` process instead of returning a normal, reportable error.

Separately, the shared HTTP transport (`nlp.post`) ignored response status codes entirely, so a non-2xx response with a technically-valid JSON body (e.g. `500 {"sents":[]}`) was silently decoded as a successful, empty result instead of a failure.

## Fix

- `post` now rejects any non-2xx response with a real error instead of decoding its body.
- `TextToTokens` returns an error instead of panicking on a failed `/tag` request. `TextToContext` and `runTag` (the `tag` CLI command) are updated to thread that error through.
- `Info.Compute` returns an error instead of panicking on a failed `/segment` request. Its existing caller, `lintProse`, already wraps any error `Compute` returns as `core.NewE100("NLP.Compute", err)`, that handling was already in place, only unreachable because `Compute` couldn't previously return an error on this path.

Each fix has its own regression test, both for `post`'s status check directly and for the panic-to-error behavior of each affected function.

## Testing

Two `internal/e2e` cases cover the user-visible behavior change end to end, each verified against the actual pre-fix binary (built at the parent commit) to confirm it currently panics there before confirming it passes cleanly here:

- `checks/remote-segment-endpoint-failure-reports-cleanly`: a lint run that hits the `Info.Compute` panic path (a `sequence` rule, non-English `Lang`, `NLPEndpoint` pointed at a closed local port for a deterministic, dependency-free connection-refused failure).
- `cli/tag-reports-endpoint-failure`: the `vale tag` CLI command hitting the same kind of failure via `TextToTokens`/`TextToContext`.

## Related

Found during review of #1162, where these were originally bundled. Split out here as an independent, single-concern fix per the smaller-PR preference from #938.

### DIFF
--- a/cmd/vale/command.go
+++ b/cmd/vale/command.go
@@ -258,8 +258,11 @@ func runTag(args []string, _ *core.CLIFlags) error {
 		return err
 	}
 
-	out := core.TextToContext(
+	out, err := core.TextToContext(
 		string(text), &nlp.Info{Lang: args[1], Endpoint: args[2]})
+	if err != nil {
+		return err
+	}
 
 	return printJSON(out)
 }
--- a/internal/core/util.go
+++ b/internal/core/util.go
@@ -265,14 +265,19 @@ func SplitLines(data []byte, atEOF bool) (adv int, token []byte, err error) { //
 	return 0, nil, nil
 }
 
-func TextToContext(text string, meta *nlp.Info) []nlp.TaggedWord {
+func TextToContext(text string, meta *nlp.Info) ([]nlp.TaggedWord, error) {
 	context := []nlp.TaggedWord{}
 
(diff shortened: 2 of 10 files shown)


jdkato:
`NLPEndpoint` is an undocumented experiment, referenced only in #356. There's a lot I don't like about its current implementation and I don't intend to support it in its current state.

PR #1168, opened 2026-09-02
chore: remove unused Tokens convenience wrappers

theredspoon:
## Problem

`File.Tokens` and `TokenCache.Tokens` are convenience wrappers around `TokensWith` (with an empty model name) that have no callers anywhere in the codebase. `TokensWith` itself is what every check type actually uses.

## Fix

Removed both. No behavior change.

## Testing

Full repo suite and `-race` are clean.

## Related

Found during review of #1162. Split out here as an independent cleanup per the smaller-PR preference from #938.

### DIFF
--- a/internal/core/file.go
+++ b/internal/core/file.go
@@ -200,14 +200,8 @@ func NewFile(src string, config *Config) (*File, error) {
 	return &file, nil
 }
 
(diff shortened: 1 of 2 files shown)


theredspoon:
Closing this as it's a standalone dead-code removal, which the new CONTRIBUTING.md guidelines say gets closed.

PR #1169, opened 2026-09-02
fix: a sequence rule with a negated scope double-reports every match

theredspoon:
## Problem

A `sequence` rule with a bare negated scope (`scope: ~list`) reaches `Run` twice for the same real match: once via the whole-block copy of the text, once more via its own `paragraph.*` wrapper.

`File.AddAlert`'s own line/span dedup absorbs this before it reaches a real `vale` run's printed output. A user running `vale` today sees the correct single alert.

The bug is still worth fixing directly. It wastes a full extra match-and-tag pass per matching document. `Scope.Matches` also has a real guarantee: a `sequence` rule declaring any scope should only ever receive single-sentence blocks. This bug silently breaks that guarantee for a negated scope. Other code that relies on it holding, a later optimization for `sequence.Run`, say, can't do so safely while this stays broken.

## Root cause

`sentenceScope`'s negation branch was a no-op: `~list` narrowed to `~list`, itself, via `strings.CutPrefix(s, "~")` re-adding the same prefix it had just stripped.

A negated term never mentions `sentence`, so `asksForSentence` (`scope.go`) then skips every `sentence.*` fragment block for such a rule. `Scope.Matches` falls back to matching both the whole-block copy and its own `paragraph.*` wrapper, since neither selector was ever narrowed to ask for a sentence fragment specifically.

## Fix

The negation branch now AND-s `sentence` in front of the term instead of leaving it untouched, so `~list` narrows to `sentence&~list`, meaning sentences outside a list, the same as every other declared scope already does.

## Testing

- `TestSentenceScope`: two updated/added rows pinning the corrected narrowing (`~list` and a chained `~list&text`).
- `TestSequenceNegatedScopeDoesNotDoubleReport`: dispatches through the real pipeline, `Info.Compute` and `Scope.Matches`, not a hand-built block. It collects `Run`'s own return value directly, which skips `AddAlert`'s dedup entirely, so the double-dispatch is actually observable here. Verified against the pre-fix behavior directly: 2 alerts for one real match before the fix, 1 after.

No `internal/e2e` case: confirmed empirically that a real `vale` run's output is identical before and after this fix (`AddAlert`'s dedup already absorbs the duplicate), so an end-to-end case here would pass either way and prove nothing.

## Related

Found during review of #1162. Split out here as an independent, single-concern fix, alongside #1167 and #1168 at the time, both since closed for unrelated reasons.

### DIFF
--- a/internal/check/sequence.go
+++ b/internal/check/sequence.go
@@ -611,11 +611,15 @@ func sentenceScope(declared []string) []string {
 			scopes = append(scopes, s)
 			continue
 		}
-		// Negation applies to the part being excluded, so it stays in front:
-		// `~list` narrows to sentences outside a list, not to something
-		// outside `sentence.list`.
-		if neg, found := strings.CutPrefix(s, "~"); found {
-			scopes = append(scopes, "~"+neg)
+		// A bare negated term names only what to exclude and never mentions
+		// `sentence` itself, so asksForSentence (scope.go) skips every
+		// `sentence.*` fragment block for it and the rule matched the whole
+		// unsegmented block instead -- narrowing to `~list` alone was a
+		// no-op. AND-ing `sentence` in front keeps the exclusion and still
+		// narrows: `sentence&~list` is "sentences outside a list", which is
+		// what the rule actually needs.
+		if strings.HasPrefix(s, "~") {
+			scopes = append(scopes, "sentence&"+s)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


theredspoon:
> I have read the CLA and I agree to it

I have read the CLA and I agree to it.

theredspoon:
recheck

theredspoon:

PR #1170, opened 2026-09-02
perf: skip redundant re-segmentation for sentence blocks in sequence.Run

theredspoon:
**Depends on #1162.** Stacked on top of it, so the diff will look larger than the actual change until #1162 merges into `v3` and this rebases. Click into this branch's single commit on the **Commits** tab to see this PR's actual diff.

## Problem

`Sequence.Run` calls `f.Sentences(blk.Text)` unconditionally, even for a plain rule whose declared scope narrows to `sentence`, meaning `blk` is already exactly one sentence a segmenter produced. Re-segmenting it is redundant, and under a configured non-English remote NLP endpoint, wastes a full HTTP round-trip per dispatched sentence.

Trusting a rule's declared scope to justify skipping that call is unsafe on its own: scope-narrowing has had real bugs before. The fix here gates the skip on a structural fact about the block itself, not on what any rule declared.

## Fix

`Block.IsSentence()` reports whether a block's scope carries the `sentence.` prefix that only the segmentation loop ever produces during block construction. `Run` skips `f.Sentences` exactly when `blk.IsSentence()` is true. `Scope.Matches` now reads the same helper for its own, unrelated `sentence.` check, so the two readings of "is this a sentence block" can't drift apart.

<details>
<summary>Why this is safe where a naive "skip when the rule has no threshold" would not be</summary>

Re-segmenting one segmenter's own output with that same segmenter can only reproduce it. The call is skipped exactly when it is truly redundant, not merely when a rule claims not to need it. A rule's own declared scope can be wrong: narrowing logic is code, and code has bugs. A block's actual constructed scope reflects how it was really built. Every block that isn't structurally known to be one sentence still gets the real call, whatever the rule declared.

</details>

## Testing

- `TestScopeMatchesOnlyHandsSequenceSentenceBlocks`: the invariant this design relies on, over the whole scope grammar.
- `TestIsSentence`: contract test pinning `Block.IsSentence` against exactly what block construction produces.
- `TestSequencePlainRuleAddsNoSegmentCallsBeyondCompute`: proves the round-trip savings by real request count against a mocked endpoint, and separately proves a real same-sentence match still resolves correctly under the gated path (verified this assertion has teeth by breaking the gate's fallback and confirming it fails).
- `TestSequenceRoundTripSkipDoesNotReintroduceNegatedScopeFalsePositive`: confirms this optimization composes correctly with the scope-narrowing fix it depends on for negated scopes.

No `internal/e2e` case. This changes network-call volume under a remote endpoint. Lint output stays the same for every correctly-configured rule, so a `vale` run's printed results show nothing different here.

## Performance

`BenchmarkSequenceRunPlainRuleSentenceBlock` measures the local, built-in-tagger cost: the map lookup this change avoids. Compared against the commit immediately before this one with `benchstat` (`n=6`):

```text
                                    │   before    │              after                 │
                                    │   sec/op    │    sec/op     vs base               │
SequenceRunPlainRuleSentenceBlock-11  1.737µ ± 1%    1.729µ ± 0%   ~ (p=0.132 n=6)

                                    │   before    │              after                 │
                                    │    B/op     │     B/op      vs base               │
SequenceRunPlainRuleSentenceBlock-11 1.814Ki ± 0%   1.814Ki ± 0%   ~ (p=1.000 n=6)

                                    │   before    │              after                 │
                                    │ allocs/op   │  allocs/op    vs base               │
SequenceRunPlainRuleSentenceBlock-11   33.00 ± 0%     33.00 ± 0%   ~ (p=1.000 n=6)
```

No measurable difference locally, as expected: the actual savings this PR targets is the eliminated remote round-trip, which the call-count test above proves directly rather than by wall-clock benchmark.

## Related

Follow-up to #1162, found during its review.

### DIFF
--- a/cmd/vale/command.go
+++ b/cmd/vale/command.go
@@ -258,8 +258,11 @@ func runTag(args []string, _ *core.CLIFlags) error {
 		return err
 	}
 
-	out := core.TextToContext(
+	out, err := core.TextToContext(
 		string(text), &nlp.Info{Lang: args[1], Endpoint: args[2]})
+	if err != nil {
+		return err
+	}
 
 	return printJSON(out)
 }
--- a/internal/check/scope.go
+++ b/internal/check/scope.go
@@ -110,7 +110,7 @@ func (s Scope) Matches(blk nlp.Block) bool {
 	// parent block too -- and that is the copy such a rule must see: a
 	// `scope: heading` rule reading a heading one fragment at a time reported
 	// `a.` as a whole heading (#1150).
-	fragment := strings.HasPrefix(blk.Scope, "sentence.")
+	fragment := blk.IsSentence()
 
 	for _, sel := range s.Selectors {
 		if fragment && !asksForSentence(sel) {
--- a/internal/check/sequence.go
+++ b/internal/check/sequence.go
@@ -609,11 +609,15 @@ func sentenceScope(declared []string) []string {
 			scopes = append(scopes, s)
 			continue
 		}
-		// Negation applies to the part being excluded, so it stays in front:
-		// `~list` narrows to sentences outside a list, not to something
-		// outside `sentence.list`.
-		if neg, found := strings.CutPrefix(s, "~"); found {
-			scopes = append(scopes, "~"+neg)
+		// A bare negated term names only what to exclude and never mentions
+		// `sentence` itself, so asksForSentence (scope.go) skips every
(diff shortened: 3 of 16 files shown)


theredspoon:
Closing this one, it's on the NLPEndpoint path, which is out of scope per CONTRIBUTING.md now.

PR #1183, opened 2026-09-19
feat(code): add Kotlin comment extraction

### LINKED ISSUE #1125: Add Kotlin support or support / document adding a Code type in vale.ini
### Check for existing issues

- [x] Completed

### Describe the feature

I would like to use vale to lint Kotlin (.kt, .kts) Kdoc (https://kotlinlang.org/docs/kotlin-doc.html) documentation comments.

Not sure what the best path forward would be. 

Does vale currently support adding a code format in the vale.ini? 
Would it be easier to add Kotlin support directly or make vale more configurable?

anandghegde:
Closes #1125.

Adds Kotlin (`.kt`, `.kts`) to the tree-sitter comment extractors, following the same layout as the Elixir support in #1151. It extracts `//` line comments and `/* */` block comments, KDoc `/** */` included (the grammar parses KDoc as an ordinary `multiline_comment`). The grammar is already vendored in go-tree-sitter, so there's no new dependency.

Two small things beyond the new language:

- The padding lists `/**` as a marker. With plain `cStyle`, an alert in a one-line KDoc comment (`/** ... */`) lands two columns short. JavaScript has the same problem today (`/** foo */` in a `.js` file), which I've left alone here.
- `TestComments` now pairs `in/N.ext` with `out/N.json` by file name instead of by position. `10.kt` sorts before `2.py`, so the positional pairing broke as soon as there were more than ten fixtures.

Tests:

- `testdata/comments/{in,out}/10.kt`, `11.kts`: line-comment runs, a KDoc block with a tag, an inline block comment, a trailing comment, a one-line KDoc, and strings containing `TODO` and `//` that must not be extracted.
- `testdata/e2e/lint.yaml`: a `kotlin` case against `testdata/fixtures/formats/test.kt`, with all eight columns checked.

`make test` passes locally (the converter-dependent cases were skipped). I also ran a built `vale` with a one-word existence rule on a `.kt` and a `.kts` file: it flagged the word only in line, block and KDoc comments, not in identifiers or strings, and the line/column numbers were right.

Assisted-by: Claude Code


anandghegde:

